PR #43, opened 2015-09-10
Jwt parser test fix

dogeared:
- Fixed broken test
- added in fail() statements inside each try that was missing one

### DIFF
--- a/src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
+++ b/src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
@@ -167,7 +167,8 @@ class JwtParserTest {
         String compact = Jwts.builder().setSubject('Joe').setExpiration(exp).compact();
 
         try {
-            Jwts.parser().parse(compact);
+            Jwts.parser().parse(compact)
+            fail()
         } catch (ExpiredJwtException e) {
             assertTrue e.getMessage().startsWith('JWT expired at ')
         }
@@ -181,7 +182,8 @@ class JwtParserTest {
         String compact = Jwts.builder().setSubject('Joe').setNotBefore(nbf).compact();
 
         try {
-            Jwts.parser().parse(compact);
+            Jwts.parser().parse(compact)
+            fail()
         } catch (PrematureJwtException e) {
             assertTrue e.getMessage().startsWith('JWT must not be accepted before ')
         }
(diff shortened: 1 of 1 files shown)


lhazlewood:
There's nothing wrong w/ semi-colons in Groovy ;)

src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
```diff
@@ -129,14 +129,14 @@ class JwtParserTest {
 
         String payload = '{"subject":"Joe"}'
 
-        String badSig = ";aklsjdf;kajsd;fkjas;dklfj"
+        String badSig = "aklsjdfkajsdfkjasdklfj"
```
lhazlewood: I'm not sure you intended this.  Gotta be careful with 'replace all'

dogeared: oh, snap! I will fix that

dogeared:
true about semicolons, but it was inconsistent in the file

src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
```diff
@@ -1,7 +1,7 @@
 /*
  * Copyright (C) 2014 jsonwebtoken.io
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Apache License, Version 2.0 (the "License")
```
lhazlewood: This is standard Apache License header - it shouldn't have been removed.  For example: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/ide-configuration/LICENSE-ASL-2.0-HEADER.txt

src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
```diff
@@ -502,7 +512,8 @@ class JwtParserTest {
         String compact = Jwts.builder().setSubject(subject).signWith(SignatureAlgorithm.HS256, key).compact()
 
         try {
-            Jwts.parser().setSigningKey(key).parseClaimsJws(compact);
+            Jwts.parser().setSigningKey(key).parsePlaintextJws(compact)
+            fail()
```
dogeared: This is the test that was actually busted, but we didn't know because of the missing `fail()`

lhazlewood:
Hrm - I'm a bit uncomfortable with the blanket replace all - it seems some things that were altered either didn't need to be or shouldn't have been.

lhazlewood:
Question - can you have one PR for the test fix and another for the semi-colon changes?  It's easier to see the real changes vs 'nosiy' syntax changes.

dogeared:
Sure. I've fixed all the spurious semicolon replacements, but I'll put it up as two PRs

dogeared:
This is now just the test fixes

src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
lhazlewood: thanks!

Issue #42, opened 2015-09-09
Enable remaining pro-active assertions when parsing a JWT


lhazlewood:
`exp` and `nbf` are already asserted automatically.  Also allow building a parser that expects values for the following:

* `iss`
* `aud`
* `sub`
* `iat`
* `jti`

Ensure the parser method names use the more friendly names - e.g. 

```java
parserBuilder.setId("jtiValue").setSubject("sub value"). ...etc... .parse(compact);
```

lhazlewood:
Implementation complete.  Will be released in 0.6.

PR #46, opened 2015-09-10
Issue 42 assert claims

dogeared:
Addresses https://github.com/jwtk/jjwt/issues/42

### DIFF
--- a/src/main/java/io/jsonwebtoken/BadAudienceJwtException.java
+++ b/src/main/java/io/jsonwebtoken/BadAudienceJwtException.java
@@ -1,0 +1,11 @@
+package io.jsonwebtoken;
+
+public class BadAudienceJwtException extends ClaimJwtException {
+    public BadAudienceJwtException(Header header, Claims claims, String message) {
+        super(header, claims, message);
+    }
+
+    public BadAudienceJwtException(Header header, Claims claims, String message, Throwable cause) {
+        super(header, claims, message, cause);
+    }
+}
--- a/src/main/java/io/jsonwebtoken/JwtParser.java
+++ b/src/main/java/io/jsonwebtoken/JwtParser.java
@@ -63,6 +63,16 @@ public interface JwtParser {
      */
     JwtParser setIssuedAt(Date issuedAt);
 
+
+    /**
+     * Sets the expected value for the jti registered claim. If the actual content of the JWT does not match
+     * (including if it's empty), then an exception will be thrown
+     *
+     * @param id the string representing the expected value for the jti (jwt id) claim of the JWT
+     * @return the parser for method chaining.
+     */
+    JwtParser setId(String id);
+
     /**
      * Sets the signing key used to verify any discovered JWS digital signature.  If the specified JWT string is not
      * a JWS (no signature), this key is not used.
--- a/src/main/java/io/jsonwebtoken/JwtParser.java
+++ b/src/main/java/io/jsonwebtoken/JwtParser.java
@@ -45,6 +45,15 @@ public interface JwtParser {
     JwtParser setAudience(String audience);
 
     /**
+     * Sets the expected value for the sub registered claim. If the actual content of the JWT does not match
+     * (including if it's empty), then an exception will be thrown
+     *
+     * @param subject the string representing the expected value for the sub (subject) claim of the JWT
+     * @return the parser for method chaining.
+     */
+    JwtParser setSubject(String subject);
+
+    /**
      * Sets the signing key used to verify any discovered JWS digital signature.  If the specified JWT string is not
      * a JWS (no signature), this key is not used.
      *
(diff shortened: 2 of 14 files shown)


src/main/java/io/jsonwebtoken/BadAudienceJwtException.java
```diff
@@ -0,0 +1,11 @@
+package io.jsonwebtoken;
```
josebarrueta: Really minor thing here, current classes in the library have the Apache license in it.

src/main/java/io/jsonwebtoken/JwtParser.java
```diff
@@ -27,6 +28,52 @@
     public static final char SEPARATOR_CHAR = '.';
 
     /**
+     * Sets the expected value for the iss registered claim. If the actual content of the JWT does not match
+     * (including if it's empty), then an exception will be thrown
+     *
+     * @param issuer the string representing the expected value for the iss (issuer) claim of the JWT
+     * @return the parser for method chaining.
+     */
+    JwtParser setIssuer(String issuer);
```
josebarrueta: When read the Javadoc was kind of confusing for me, shouldn't be something like:

Sets the expected value for the iss registered claim. If set to a non-null value, the ```iss``` claim in the JWT will be validated against it....

josebarrueta: Sort of the same comment for the ```aud```, ```sub```, ```iat``` and ```jti```

lhazlewood: I think we need a `expect(String claimName, Object value)` base method that can be used to assert any claim - not just the standard ones.

Then, we can have convenience methods for the standard ones that delegate to that method.

After thinking about it, I'd probably also call these methods `expectIssuer`, `expectAudience`... etc.  The reason is that you're not really setting properties of the parser itself (which is what most setter methods are for - configuring the invoked object), but telling the parser to expect certain values.  

Thoughts?

src/main/java/io/jsonwebtoken/BadAudienceJwtException.java
```diff
@@ -0,0 +1,11 @@
...
+
+public class BadAudienceJwtException extends ClaimJwtException {
```
lhazlewood: I think these should be more generic, instead of an exception per condition.

Perhaps something like this:

```java
public class InvalidClaimException extends ClaimJwtException { ... }

//when a value is expected, but one isn't provided in the JWT:
public class MissingClaimException extends InvalidClaimException { ... }

//when a value is expected and one is provided, but the provided value doesn't match the expected value:
public class IncorrectClaimException extends InvalidClaimException {...}
```
(ClaimJwtException should probably be renamed to ClaimException, but that's another issue)

Thoughts?

src/main/java/io/jsonwebtoken/JwtParser.java
dogeared: Yes, I agree. I think that makes a clear delineation between setting and expecting.

src/main/java/io/jsonwebtoken/BadAudienceJwtException.java
dogeared: Nice! Right now there's no distinction between missing and unexpected (which I like better than `incorrect` as it matches with `expect` from the parser)

lhazlewood: I thought about `UnexpectedClaimException` but avoided it - I think it can too easily confuse people because of claims that exist, and are legal, but aren't specifically listed when building the parser.

For example, I don't 'expect' an IAT field, but it is there.  Unexpected might be seen as "a claim exists that I didn't expect to see" as opposed to "the value for a required claim is not the correct value"

dogeared: hmm... Maybe it should be `ExpectedClaimException` - there was an expectation that wasn't met.

`IncorrectClaimException` is technically correct, but it doesn't match up with the `expect` and `expectIssuer`, etc. methods. Not a big showstopper for me, though - I get stuck on words sometimes ;)

lhazlewood: Hah, you and me both - I agonize over naming because I know how much it contributes to readability and intuitive APIs...

lhazlewood: Let's go w/ `IncorrectClaimException` for now - correlated names are nice, but I think we'll be fine in this case.

lhazlewood: An alternative to `expect` based on your concerns might be `require` -  `requireIssuer`, etc.  Maybe then `IncorrectClaimException` would feel better?

dogeared:
put up new pull request that addresses the comments in this one.

PR #49, opened 2015-09-12
Issue 42 assert claims new

dogeared:
Addresses #42

### DIFF
--- a/src/main/java/io/jsonwebtoken/IncorrectClaimException.java
+++ b/src/main/java/io/jsonwebtoken/IncorrectClaimException.java
@@ -16,8 +16,8 @@
 package io.jsonwebtoken;
 
 /**
- *  IncorrectClaimException is a subclass of the {@link InvalidClaimException} that is thrown after it is found that an
- *  expected claim has a value that is not expected.
+ * Exception thrown when discovering that a required claim does not equal the required value, indicating the JWT is
+ * invalid and may not be used.
  *
  * @since 0.6
  */
--- a/src/main/java/io/jsonwebtoken/IncorrectClaimException.java
+++ b/src/main/java/io/jsonwebtoken/IncorrectClaimException.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2015 jsonwebtoken.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.jsonwebtoken;
+
+/**
+ *  IncorrectClaimException is a subclass of the {@link InvalidClaimException} that is thrown after it is found that an
+ *  expected claim has a value that is not expected.
+ *
+ * @since 0.6
+ */
+public class IncorrectClaimException extends InvalidClaimException {
+    public IncorrectClaimException(Header header, Claims claims, String message) {
+        super(header, claims, message);
+    }
+
+    public IncorrectClaimException(Header header, Claims claims, String message, Throwable cause) {
+        super(header, claims, message, cause);
+    }
+}
--- a/src/main/java/io/jsonwebtoken/InvalidClaimException.java
+++ b/src/main/java/io/jsonwebtoken/InvalidClaimException.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2015 jsonwebtoken.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.jsonwebtoken;
+
+/**
+ * InvalidClaimException is a subclass of the {@link ClaimJwtException} that is thrown after a validation of an JTW claim failed.
+ *
+ * @since 0.6
+ */
+public class InvalidClaimException extends ClaimJwtException {
+    private String claimName;
+    private Object claimValue;
+
+    protected InvalidClaimException(Header header, Claims claims, String message) {
+        super(header, claims, message);
+    }
+
+    protected InvalidClaimException(Header header, Claims claims, String message, Throwable cause) {
+        super(header, claims, message, cause);
+    }
+
+    public String getClaimName() {
+        return claimName;
+    }
+
+    public void setClaimName(String claimName) {
+        this.claimName = claimName;
+    }
+
+    public Object getClaimValue() {
+        return claimValue;
+    }
+
+    public void setClaimValue(Object claimValue) {
+        this.claimValue = claimValue;
+    }
+}
--- a/src/main/java/io/jsonwebtoken/InvalidClaimException.java
+++ b/src/main/java/io/jsonwebtoken/InvalidClaimException.java
@@ -16,7 +16,11 @@
 package io.jsonwebtoken;
 
 /**
- * InvalidClaimException is a subclass of the {@link ClaimJwtException} that is thrown after a validation of an JTW claim failed.
+ * Exception indicating a parsed claim is invalid in some way.  Subclasses reflect the specific
+ * reason the claim is invalid.
+ *
+ * @see IncorrectClaimException
+ * @see MissingClaimException
  *
  * @since 0.6
  */
--- a/src/main/java/io/jsonwebtoken/JwtParser.java
+++ b/src/main/java/io/jsonwebtoken/JwtParser.java
@@ -16,6 +16,7 @@
 package io.jsonwebtoken;
 
 import java.security.Key;
+import java.util.Date;
 
 /**
  * A parser for reading JWT strings, used to convert them into a {@link Jwt} object representing the expanded JWT.
@@ -25,6 +26,14 @@ import java.security.Key;
 public interface JwtParser {
 
     public static final char SEPARATOR_CHAR = '.';
+
+    /**
+     * Sets an expected value for the issuedAt claim.
+     *
+     * @param issuedAt
+     * @return the parser for method chaining.
+     */
+    JwtParser expectIssuedAt(Date issuedAt);
 
     /**
      * Sets an expected value for any given claim name.
--- a/src/main/java/io/jsonwebtoken/JwtParser.java
+++ b/src/main/java/io/jsonwebtoken/JwtParser.java
@@ -28,81 +28,99 @@ public interface JwtParser {
     public static final char SEPARATOR_CHAR = '.';
 
     /**
-     * Ensures that the specified {@code jti} value is present when parsing the JWT.  If not present,
-     * an exception will be thrown indicating that the JWT is invalid and may not be used.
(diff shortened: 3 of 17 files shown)


src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -58,6 +62,54 @@
 
     private SigningKeyResolver signingKeyResolver;
 
+    Claims expectedClaims = new DefaultClaims();
+
```
dogeared: Is this cheating? Date gets handled differently internally, so I did the null check and then used the specific setter.

All the other claims are String based, so I just call `expect`. This makes it so I don't have to do a null check for each type.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -309,6 +363,47 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
...
     }
 
+    private void validateExpectedClaims(Header header, Claims claims) {
+        for (String expectedClaimName : expectedClaims.keySet()) {
+            Object expectedClaimValue;
+            Object actualClaimValue;
+
+            // since issued at is a date, call the specific method
+            // other methods deal with strings and the more
+            // general method can be used
```
dogeared: Similar to the above, there's a check to see if it's an `issedAt` check. All other types are handled in the `else`

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -58,6 +62,54 @@
...
+    @Override
+    public JwtParser expectId(String id) {
+        expect(Claims.ID, id);
+
+        return this;
+    }
+
+    @Override
+    public JwtParser expect(String claimName, Object value) {
+        if (claimName != null && claimName.length() > 0 && value != null) {
```
lhazlewood: This should throw an exception if claimName is null or length of zero and should throw an exception if the value is null - either scenario means the caller is doing something wrong (invalid API usage).

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
lhazlewood: I don't think this is necessary - we should invoke the delegate method immediately (like the others) and the delegate should have date-specific logic if necessary.

Users can call the delegate method with a custom claim name, and if that value is a Date, we still have to handle that correctly.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -58,6 +62,54 @@
 
     private SigningKeyResolver signingKeyResolver;
 
+    Claims expectedClaims = new DefaultClaims();
+
+    @Override
+    public JwtParser expectIssuedAt(Date issuedAt) {
+        if (issuedAt != null) {
```
lhazlewood: no null check - just invoke the delegate directly (and null values are not allowed for required claim checks as indicated in the delegate method comments below).

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
lhazlewood: Can't do this - users can expect a custom data field, and we have to handle that correctly as well.  i.e. we shouldn't have special logic for just one date field - any number of date fields could be present.

src/main/java/io/jsonwebtoken/IncorrectClaimException.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.jsonwebtoken;
+
+/**
+ *  IncorrectClaimException is a subclass of the {@link InvalidClaimException} that is thrown after it is found that an
```
lhazlewood: That this is a subclass is obvious just by the nature of the generated JavaDoc and a little heavy-handed for documentation IMO.  I recommend keeping it direct and simple:

> Exception thrown when parsing a required claim and the value does not equal the required value.

src/main/java/io/jsonwebtoken/InvalidClaimException.java
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.jsonwebtoken;
+
+/**
+ * InvalidClaimException is a subclass of the {@link ClaimJwtException} that is thrown after a validation of an JTW claim failed.
```
lhazlewood:
```java
/**
 * Exception indicating a parsed claim is invalid in some way.  Subclasses reflect the specific 
 * reason the claim is invalid.
 *
 * @see IncorrectClaimException
 * @see MissingClaimException
 */
```

For abstract classes or intermediate classes in public APIs, I try to show or link to other examples that give the class more contextual relevance.  Using the `@see` jdoc annotation is a simple/useful way to do this.

src/main/java/io/jsonwebtoken/MissingClaimException.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.jsonwebtoken;
+
+/**
+ *  MissingClaimException is a subclass of the {@link InvalidClaimException} that is thrown after it is found that an
```
lhazlewood:
```java
/**
 * Exception thrown when discovering that a required claim is not present, indicating the JWT is invalid and may not be used.
 */
```

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -58,6 +62,54 @@
...
+
+    @Override
+    public JwtParser expectId(String id) {
+        expect(Claims.ID, id);
+
+        return this;
+    }
+
+    @Override
+    public JwtParser expect(String claimName, Object value) {
```
lhazlewood: I think this and the others should be renamed to `require` - it feels a little more direct and authoritative such that a failure will occur if missing.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
dogeared: Right. I get that. I was just trying to avoid having a ton of branching logic. I was thinking of having the registered claim types as an enum instead of constants so we could `switch` on it or using reflection, but that seemed like overkill.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
dogeared: makes sense

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
dogeared: got it.

lhazlewood:
Ping!  Just a friendly bump to see if you've made any changes based on our last review.  Cheers!

dogeared:
This is nearly there. There's an issue with using the generic `require` method and dates.

This test fails:

```
    @Test
    void testParseExpectedCustomDate_Success() {
        def aDate = new Date(System.currentTimeMillis())

        byte[] key = randomKey()

        String compact = Jwts.builder().signWith(SignatureAlgorithm.HS256, key).
            claim("aDate", aDate).
            compact()

        Jwt<Header,Claims> jwt = Jwts.parser().setSigningKey(key).
            expect("aDate", aDate).
            parseClaimsJws(compact)

        assertEquals jwt.getBody().get("aDate"), aDate
    }
```

The call to `compact` converts the `Date` object to a `long`. The call to `get` returns the long. There's no internal knowledge that it was ever a `Date` object.

Could use some feedback on the best approach here.

lhazlewood:
I've seen a convenient pattern that solves this in other frameworks, and we can do something similar.  For example, this can be solved by adding the following method to the Claims interface:

```java
T get(String name, Class<T> expectedType);
```

The second argument would be used to assert the value returned is null or adheres to the specified type.  Then you could call this:

```java
Date date = claims.get("aDate", Date.class);
```

The implementation can check the 2nd argument - if it is a Date, we know to perform the seconds-not-milliseconds-since-epoc --> date conversion.

If it is not null or a Date, we just assert that the value obtained is assignable:

```java
Object value = get(name);
if (value == null) {
    return null;
}
If !(expectedType.isInstance(value)) {
    //throw a custom JWT exception here, e.g. UnexpectedTypeException or something similar
}

return expectedType.cast(value);
```

src/main/java/io/jsonwebtoken/JwtParser.java
```diff
@@ -27,6 +28,79 @@
     public static final char SEPARATOR_CHAR = '.';
 
     /**
+     * Sets an expected value for the jti claim.
```
lhazlewood: Let's change these comments to reflect the 'require' nomenclature:

```java
/**
 * Ensures that the specified {@code jti} value is present when parsing the JWT.  If not present, 
 * an exception will be thrown indicating that the JWT is invalid and may not be used.
 */
```

src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
```diff
@@ -106,4 +107,20 @@ public Claims setId(String jti) {
...
         return this;
     }
+
+    @Override
+    public <T> T get(String claimName, Class<T> requiredType) {
+        Object value = get(claimName);
+        if (value == null) { return null; }
+
+        if (requiredType == Date.class && value instanceof Long) {
+            value = new Date((Long)value);
```
lhazlewood: I don't think this will work if the claim name requested is a standard JWT date field, because that value is required to be in seconds (not millis) since epoch.

In other words, if I do:

```java
claims.get("iat", Date.class);
```

It should work.

In fact, the other get* methods that return a date should probably delegate to this method so the behavior is guaranteed.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -309,6 +375,49 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
...
     }
 
+    private void validateExpectedClaims(Header header, Claims claims) {
+        for (String expectedClaimName : expectedClaims.keySet()) {
+
+            Object expectedClaimValue = expectedClaims.get(expectedClaimName);
+            Object actualClaimValue = claims.get(expectedClaimName);
+
+            if (Claims.ISSUED_AT.equals(expectedClaimName)) {
+                expectedClaimValue = expectedClaims.getIssuedAt();
```
lhazlewood: I don't think we need these 3 separate if statements if the get(name,date) method is adjusted as mentioned above.

src/main/java/io/jsonwebtoken/JwtParser.java
```diff
@@ -27,6 +28,85 @@
     public static final char SEPARATOR_CHAR = '.';
 
     /**
+     * Ensures that the specified {@code jti} value is present when parsing the JWT.  If not present,
```
lhazlewood: Hrm - sorry about the ambiguity here.

How about this:

```java
/**
 * Ensures that the specified {@code jti} exists in the parsed JWT.  If missing or if the parsed 
 * value does not equal the specified value, an exception will be thrown indicating that the 
 * JWT is invalid and may not be used.
 */
```

src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
```diff
@@ -65,7 +66,7 @@ public Claims setAudience(String aud) {
 
     @Override
     public Date getExpiration() {
-        return getDate(Claims.EXPIRATION);
+        return get(Claims.EXPIRATION, Date.class);
```
lhazlewood: Nice!

PR #54, opened 2015-09-24
Issue 52

josebarrueta:


lhazlewood:
@josebarrueta I think we need to update the JavaDoc on all compression-related methods on the builder:  we need to be *super* clear that this is not an official JWT specification feature and that other JWT parsers other than JJWT are likely to not support compression.  I.e. "Only use compression if the token will be consumed by JJWT later" kind of thing...

josebarrueta:
Hi @lhazlewood, just added a commit improving the javadocs for the new JWT compression,stating that is not part of the spec, your feedback is much appreciated.

Issue #48, opened 2015-09-11
Add build Jwt to JwtBuilder 


danilobuerger:
`JwtBuilder` should have the ability to actually build a `Jwt` object instead of only the compact `String` representation.

I would 
- add `Jwt build();` to `JwtBuilder`
- add `String compact();` to `Jwt`
- move all functionality (minus return `String`) from `DefaultJwtBuilder.compact()` to `DefaultJwtBuilder.build()`
- implement `DefaultJwt.compact()` and `DefaultJws.compact()` to return the compact `String`
- make `DefaultJwtBuilder.compact()` call `return build().compact();` to not break api compatibility

If a core contributor could tell me if that would be accepted as a PR, i would do it. Otherwise, i could also implement another solution to this if outlined.

lhazlewood:
What is the use case where a JWT object needs to be built instead of the compact string?  I'm just trying to understand the desired change since almost everyone only uses JWTs to 1) build the compact string, and then later 2) parse a compact string

danilobuerger:
@lhazlewood I have an endpoint which spits out a JWT token. This token is later used by clients to auth. I want to make it easier for clients so that they don't necessarily have to parse the token. Thats why i am return the claims in my response to:

```
{
  "claims": {
    "sub": "...",
    ...
  },
  "token": "compact"
}
```

If we had a ``build`` i could do:

```
final Jwt jwt = builder.build();
response.put("claims", jwt.getBody());
response.put("token", jwt.compact());
````

Which would be awesome :dancer: 

Since we are not breaking API, and just adding another two public methods, and i would write it, i guess there is no harm to it even if you don't see my use case as a valid one. People will always do crazy stuff with libraries :tada:

lhazlewood:
This is helpful, thanks!

I think the sentiment of the request definitely makes sense.  The only thing I'm currently debating is the notion of having `jwt.compact()` - this requires the JWT implementation know about JSON/JWT compaction logic, which feels a little too 'coupled' to me at the moment (not sure though - just thinking out loud).

In other words, in the spirit of the Single Responsibility Principle, it seems as if formatting of a JWT into a compact representation should be a separate concern from the JWT implementation itself.  Or maybe given that compaction is so core to JWT itself, that you just ignore SRP and do it anyway?

Thoughts?

danilobuerger:
[...]

PR #56, opened 2015-10-08
Introduced JwtBuilder.build method to allow creation of Jwt objects

danilobuerger:
Relates to #48.

With this PR ``compact`` would start using ``build`` to assemble the serialized jwt. A disadvantage of this is if i want to ``build`` a jwt and then afterwards ``compact`` it, as it would then call ``build`` again. I still think if would be more of an advantage to move ``compact`` to the ``Jwt``.


lhazlewood:
@danilobuerger Apparently there is a conflict w/ this PR and master - could you updated it when you get a chance please?

danilobuerger:
@lhazlewood amended

Issue #66, opened 2015-11-18
custom claims


leantony:
How do I add a map of custom claims to the jwt ? For example, if I wanted to add a user's role
Currently iam doing it like this

``` java
return Jwts.builder()
                .setSubject(subject)
                .setIssuer(issuer)
                .setId(UUID.randomUUID().toString())
                .setExpiration(setExpiry(expiresIn))
                .setClaims(claims)
                .signWith(this.Algo, this.getEncodedSecret())
                .compact();
```

This is just a portion of a function. Claims is variable of Map<String, Object> type. Iam using this as the custom claims, but when I do this, my token is generated without an expiry date. What am I doing wrong?

lhazlewood:
Try setting the claims before you set any other field:

```java
return Jwts.builder()
                .setClaims(claims)
                .setSubject(subject)
                .setIssuer(issuer)
                .setId(UUID.randomUUID().toString())
                .setExpiration(setExpiry(expiresIn))
                .signWith(this.Algo, this.getEncodedSecret())
                .compact();
```

I suspect that setClaims is overwriting the other claims that have already been set.

If this works, let me know - I'll leave the ticket open to copy the map key/value pairs into the existing map (rather than replace the existing map).

leantony:
Yes. I noticed it was overwriting the other claims. So, had to do this:
```java
return Jwts.builder()
                .claim("username", "the user's name")
                .setIssuer(issuer)
                .setExpiration(setExpiry(expiresIn))
                .setId(UUID.randomUUID().toString())
                .setSubject(subject)
                .signWith(this.Algo, this.getEncodedSecret())
                .compact();
```
I thought SetClaims() would take in my custom claims, but on reading through the source code, I saw that it wasn't doing that. But with my above method, I have to repeat the claim('key', 'value') portion, for any custom claim I have to add

lhazlewood:
I've thought about this for a while now.

Per standard/idiomatic Java setter methods (where state is replaced with what is currently being specified) setClaims *should* replace any current claims that might be set.

That said, I think we should add a `addClaims` method that takes an input map and appends it to the existing populated claims.

lhazlewood:
Closing as this his was resolved via #196, released in 0.8.0

PR #71, opened 2015-12-07
appending custom claims to the JWT

leantony:
Hi. With reference to issue #66, i've created a pull request that attempts to fix the issue. Mine is an Easy fix really. Please review

### DIFF
--- a/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
+++ b/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
@@ -255,16 +255,14 @@ public class DefaultJwtBuilder implements JwtBuilder {
     
     /**
      * Add custom claims to the jwt payload
-     * @param customClaims map containing custom claims
+     * @param claims map containing custom claims
      * @return JwtBuilder
      */
-    public JwtBuilder appendCustomClaims(Map<String, Object> customClaims){
-        if(customClaims.isEmpty()){
-            throw new IllegalStateException("Claims cannot be empty");
-        } else {
-            for (Map.Entry<String, Object> c : customClaims.entrySet()){
-                claim(c.getKey(), c.getValue());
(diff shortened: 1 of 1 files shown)


src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
```diff
@@ -252,7 +252,24 @@ public JwtBuilder claim(String name, Object value) {
...
         return this;
     }
+    
+    /**
+     * Add custom claims to the jwt payload
+     * @param customClaims map containing custom claims
+     * @return JwtBuilder
+     */
+    public JwtBuilder appendCustomClaims(Map<String, Object> customClaims){
+        if(customClaims.isEmpty()){
```
lhazlewood: This will throw an NPE if `customClaims` is null.  Instead, I'd do the following:

```java
Assert.notEmpty(customClaims, "Claims map cannot be null or empty.");
```

Additionally, this approach eliminates the if/else flow, which is easier to read and reduces cyclomatic complexity.  The implementation should look like this:

```java
Assert.notEmpty(customClaims, "Claims map cannot be null or empty.");

for (Map.Entry<String, Object> c : customClaims.entrySet()) {
    claim(c.getKey(), c.getValue());
}

return this;
```

Also, please change the method signature to be:

```java
public JwtBuilder addClaims(Map<String,Object> claims);
```

('add' is more aligned with Java Collections naming semantics.  It is also congruently named with the `setClaims` method)

Can you please update the PR accordingly?  Thank you!

leantony:
I haven't written any tests for the PR. Guess that's why the automated tests are failing

lhazlewood:
Yep - we enforce 100% coverage where possible. :)

lhazlewood:
@leantony do you want to add a test case for this?  If not, we should close this PR...

Issue #61, opened 2015-10-27
JWT NBF and clock drift


timchaffee:
4.1.4.  "exp" (Expiration Time) Claim

   The "exp" (expiration time) claim identifies the expiration time on
   or after which the JWT MUST NOT be accepted for processing.  The
   processing of the "exp" claim requires that the current date/time
   MUST be before the expiration date/time listed in the "exp" claim.
   **Implementers MAY provide for some small leeway, usually no more than
   a few minutes, to account for clock skew.**  Its value MUST be a number
   containing a NumericDate value. Use of this claim is OPTIONAL.

4.1.5.  "nbf" (Not Before) Claim

   The "nbf" (not before) claim identifies the time before which the JWT
   MUST NOT be accepted for processing.  The processing of the "nbf"
   claim requires that the current date/time MUST be after or equal to
   the not-before date/time listed in the "nbf" claim.  **Implementers MAY
   provide for some small leeway, usually no more than a few minutes, to
   account for clock skew.**  Its value MUST be a number containing a
   NumericDate value.  Use of this claim is OPTIONAL.

Any chance you can add the ability to set a skew/drift parameter to the exp and nbf check?

lhazlewood:
Thanks for the issue - any suggestions on how this would be specified by a JJWT user?

timchaffee:
Hmm - not sure - maybe:

Jwts.parser().setSigningKey(key).**setAllowedClockSkew(3)**.parseClaimsJws(s).getBody().getSubject().equals("Joe");

ca0abinary:
+1 jose4j has similar functionality: https://bitbucket.org/b_c/jose4j/wiki/JWT%20Examples

mrwilby:
+1 . Thinking how to do this now myself.

lhazlewood:
fixed in master thanks to @ MichaelSims.  Will be in 0.7 shortly.

PR #74, opened 2015-12-17
Expiration extension for parser

strokine:
In some cases I need to accept even expired token.
I've added a setting to the parser:
```
Jwts.parser().setExpirationExtension(2000).parse(compact)
```
which would extend the expiration time by the provided value (in milliseconds).

If you like the feature and accept the pull request, it would help us continue to use the official release instead of our fork.

Thank you,
Eugene

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
         <maven.jar.version>2.4</maven.jar.version>
         <maven.compiler.version>3.1</maven.compiler.version>
 
-        <jdk.version>1.6</jdk.version>
+        <jdk.version>1.7</jdk.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <buildNumber>${user.name}-${maven.build.timestamp}</buildNumber>
 
--- a/src/main/java/io/jsonwebtoken/JwtParser.java
+++ b/src/main/java/io/jsonwebtoken/JwtParser.java
@@ -234,7 +234,16 @@ public interface JwtParser {
      * @return the parser for method chaining.
      * @since 0.7.0
      */
-    JwtParser setExpirationExtension(long expirationExtension);
+    JwtParser setExpirationExtensionMillis(long expirationExtensionMillis);
+    
+    /**
+     * Time skew for premature and expiration timeout. 
+     * 
+     * @param driftTimeMillis time skew in milliseconds
+     * @return the parser for method chaining.
+     * @since 0.7.0
+     */
+    JwtParser setDriftTimeMillis(long driftTimeMillis);
 
     /**
      * Returns {@code true} if the specified JWT compact string represents a signed JWT (aka a 'JWS'), {@code false}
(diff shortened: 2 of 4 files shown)


lhazlewood:
Hi @strokine, thanks for the PR!  However, we have to do this work as part of the work for #61 - to support clock drift for both `nbf` and `exp`.

Would you be willing to reformulate the PR to account for both via a `allowClockDrift` builder method?  Thoughts?

strokine:
Those ideas are different. #61 is to be flexible with the possible clock difference on the servers which issue or read the token.

This pool request is for the case when you expect an expired token. For example, lets say a token issuer goes down, but token requestor cached some tokens and now is pushing those tokens to verifier. 
Settings of the verifier could be adjusted to say something like this: "accept the tokens for another hour, while we are dealing with the token issuer problem".

So, I think #61 should be implemented as well, and the clock drift should be applied even on the extended time. But this is a separate issue.

lhazlewood:
But wouldn't setting the clock drift value to be the same as your expiration time address the issue?

lhazlewood:
In other words:

```java
.allowClockDrift(expirationTimeMillis + driftTimeMillis);
```

strokine:
If I understand the idea of clock drift correctly, it adds some space to two check points:
1. premature token exception would be thrown only if the **currentTime + driftTimeMillis < tokenIssueTime**
2. expired token exception would be thrown only if the **currentTime - driftTimeMillis > tokenExpirationTime**

So, expirationTimeMillis would be applied only to case 2, like this: **currentTime - driftTimeMillis > tokenExpirationTime+expirationTimeMillis**
So both properties are needed, because they are serving different cases, and even if expirationTimeMillis is set, the premature token exception wouldn't be affected.

Unfortunately I don't have time today to add one more feature for the driftTimeMillis, but if I'll have it tomorrow, I'll update this PR or will create a new one. But even without driftTimeMillis feature, expirationTimeMillis still have right to exist.

strokine:
I've added DriftTimeMillis property as well

strokine:
@lhazlewood do you think this PR could be accepted? Just don't want to maintain my own fork. Thanks.

lhazlewood:
Sorry for the delay on this @strokine.  I do think the PR can be accepted, but I want to think through this design a bit more.  It shouldn't be long.

lhazlewood:
Closing as this was released in 0.7.0 as parser.`setAllowedClockSkewSeconds`.  The method name was chosen to match the terminology ('skew') used in the RFC specification.  You can process expired tokens with this by setting the skewSeconds to be a large enough value.  If you believe this to be in error or does not satisfy your use case, please do re-open and we can discuss further!

lhazlewood:
Also note that with 0.7.0 you can specify your own clock on the parser: `....setClock(myClock);` and this clock can return whatever time you want for the `now` method.  This allows you to avoid `nbf` and `exp` errors by specifying a time that makes validation pass - useful for tests or parsing old tokens if necessary.  PR #80 (if accepted) could be another way of bypassing validation as well...

PR #85, opened 2016-01-27
#61: Add support for clock skew to JwtParser for exp and nbf claims

MichaelSims:
I realize there is an existing pull request that references this issue, but it combines an implementation for the issue with another related but separate one. Thought there might be some value in having one that addresses this issue directly as an alternative. (This is my first pull request, my apologies if I've unwittingly committed a faux pas.)


jd-carroll:
These changes seem reasonable, what is the status of getting this committed?

lhazlewood:
@jd-carroll currently branch conflicts.  These need to be resolved before we review the PR.

MichaelSims:
@lhazlewood I rebased the branch to latest master. Thanks!

PR #95, opened 2016-02-22
Made the android dep optional in OSGi

dleblanc:
Changed the packaging type to bundle - required
by the bundle plugin.

Upgraded to the latest version of the maven
bundle plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -355,7 +355,7 @@
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
-                <version>2.3.5</version>
+                <version>3.0.1</version>
                 <extensions>true</extensions>
                 <executions>
                     <execution>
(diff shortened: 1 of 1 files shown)


dleblanc:
Also - the CompressionCodecs class is in the .impl.compression package, which isn't visible to the OSGi bundle, so we can't actually provide a compression codec to use. Perhaps either make a visible method to expose those (eg: .deflateCompression()) in a visible non-impl package, or move the CompressionCodecs to the non-impl package.

lhazlewood:
The build is failing, so that should be addressed when possible.

Also, maybe the bundle should be a separate artifact?  OSGi seems to be a special use case these days - I'm afraid not as many people use it/require it any longer, and I'm hesitant to make this a hard requirement for JJWT.

Also what problem were you having with CompressionCodecs?  How did they manifest?

dleblanc:
Marking the build as a bundle just means the manifest gets stamped out differently - it doesn't change anything for anyone who isn't using it in OSGi. Most projects (Guava, Spring, etc) do this.

The build failure seems to be related to coburtura crash (a bug I would presume), but it appears that subsequent builds are passing.

lhazlewood:
If I specify a `<classifier>` in a Maven pom as `jar`, it won't fail?

dleblanc:
Just checked - it works fine with the specifier as 'jar' too. I've updated the pull request.

dleblanc:
Just let me know if there's anything else you need on my side - the change set is now trivially small, so it should be good to merge.

PR #96, opened 2016-02-23
allow the injection of a time source

Blackbaud-MitchellMorris:
This PR allows the injection of a clock source into the JwtParser for cases where we need to control the clock used during parsing.

### DIFF
--- a/src/main/java/io/jsonwebtoken/JwtParser.java
+++ b/src/main/java/io/jsonwebtoken/JwtParser.java
@@ -125,13 +125,13 @@ public interface JwtParser {
     JwtParser require(String claimName, Object value);
 
     /**
-     * Replaces the system time-of-day call to return this fixed value instead. A {@code null} value will remove
-     * the fixed date output and return the system time-of-day instead.
+     * Replace the {@code clock} used by the parser to determine the current time-of-day to use when validating
+     * the parsed JWT. If {@code null}, will reset the behavior to use the system clock.
      *
-     * @param now the time-of-day to use for parsimg this JWT or {@code null} to use the current time-of-day
+     * @param clock a {@code Clock} object to return the time-of-day or {@code null}
      * @return the builder instance for method chaining.
      */
-    JwtParser setFixedClock(Date now);
+    JwtParser setClock(Clock clock);
 
     /**
      * Sets the signing key used to verify any discovered JWS digital signature.  If the specified JWT string is not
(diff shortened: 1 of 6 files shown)


Blackbaud-MitchellMorris:
There appears to be a build environment failure in Travis-CI for OpenJDK 7. I am unable to investigate your build environment so I have no way to proceed with debugging.

lhazlewood:
You can investigate it - it is just a Travis CI build.  Please see the `.travis.yml` at the root of the project and the [Travis documentation](https://docs.travis-ci.com/user/customizing-the-build/) accordingly.

src/main/java/io/jsonwebtoken/JwtParser.java
```diff
@@ -125,6 +125,15 @@
     JwtParser require(String claimName, Object value);
 
     /**
+     * Replaces the system time-of-day call to return this fixed value instead. A {@code null} value will remove
+     * the fixed date output and return the system time-of-day instead.
+     *
+     * @param now the time-of-day to use for parsimg this JWT or {@code null} to use the current time-of-day
+     * @return the builder instance for method chaining.
+     */
+    JwtParser setFixedClock(Date now);
```
lhazlewood: Per the wording of your own issue subject line, I think it'd be better/more robust if the user can inject their own clock:

```java
JwtParser setClock(Clock clock) {
    this.clock = clock;
}
```

Where Clock looks like this:

```java
public interface Clock {
    Date now();
}
```

And you could have a default implementation:

```java
public class FixedClock implements Clock {

    private final Date date;

    public FixedClock() {
        this(new Date());
    }
    public FixedClock(Date date) {
        this.date = date;
    }

    public Date now() {
        return this.date;
    }
}
```

And then you could call:

```java
Jwts.parser().setClock(new FixedClock(new Date()));
```

Allows for more extensible implementations and testing scenarios.

lhazlewood: And we'd need a `DefaultClock` implementation would just return `new Date();` in its `now()` implementation.

lhazlewood:
btw, I asked Travis to re-run the build to see if the failure is repeatable.  In any event, if you implement the `Clock` approach, I can accept the PR and take it from there.

Blackbaud-MitchellMorris:
Thank you for your feedback

arianvp:
Can't we switch to  java.time while we're at it? Of course it's only java8 but it does support injecting time sources wiht `java.time.Clock`

lhazlewood:
Unfortunately we can't force `java.time.Clock` as Java 8 isn't universally adopted yet: http://www.baeldung.com/java-8-adoption-march-2016

lhazlewood:
Just an update here:  I incorporated this PR into #110 so I could add some Javadoc and another test (we require 100% code coverage from tests, and I needed to add one), so I'm closing this in favor of 110.  Thanks for the PR!

Issue #78, opened 2015-12-30
Cannot use EllipticCurveSignatureValidator with Key bytes.


zemelett:
In DefaultJwtParser.parse if a Key object is not provided but key bytes is not empty a SecretKeySpec object is created for the key:
```java
key = new SecretKeySpec(keyBytes, algorithm.getJcaName());
```
If however the algorithm is ES256, ES384 or ES512, the DefaultSignatureValidatorFactory will try to use an EllipticCurveSignatureValidator.  Which will check that the key is an instance of ECPublicKey, which it won't be because of the above.

lhazlewood:
Resolved by #105

PR #105, opened 2016-03-20
Avoid potentially critical vulnerability in ECDSA signature validation

aarondav:
Quite possible we're missing something here, so please forgive if so. After seeing [this article](https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/) (see "RSA or HMAC?" section), we did a quick scan through the JJWT implementation to see if it was vulnerable. While it seems like the RSA check should work, no such check seemed to exist for ECDSA signatures.

As a result, it may be possible for users of this library to use `setSigningKey(byte[] key)` while intending to use ECDSA, but have the client alter the algorithm and signature to use HMAC with the public key as the "secret key", allowing the client to inject arbitrary payloads.

cc @thomaso-mirodin


lhazlewood:
Thanks for the issue - wouldn't it just be better to Assert.isTrue(algorithm.isHmac()) to avoid missing any potential future algorithms that might come along?

aarondav:
Done! Thanks for taking a look.

lhazlewood:
When merged, this resolves #78

PR #117, opened 2016-04-27
implement hashCode and equals in JwtMap

matzon:
Having problems implementing equals/hashcode methods of classes using headers and claims.
* Added (missing!?) equals/hashCode and call-through to AbstractMap


matzon:
Please note the Java specific test, to ensure we hit proper Java methods and not Groovy's inbuilt

lhazlewood:
Thanks for the PR!

For future reference, there is no need to have any test class be in Java.  Groovy is 100% interoperable with Java and you can call Java methods from Groovy without any problems.  If you turn on code coverage (e.g. in your IDE), you will see the JwtMap methods are called regardless of how Groovy calls them.  That's all that matters.  Cheers!

matzon:
Updated PR. I could have sworn the DefaultGroovyMethods was causing some issues - but not this time around.

src/main/java/io/jsonwebtoken/impl/JwtMap.java
```diff
@@ -155,4 +155,16 @@ public void clear() {
     public String toString() {
         return map.toString();
     }
+
+    @Override
+    public int hashCode()
+    {
```
lhazlewood: Can you please use standard Java syntax formatting?

src/main/java/io/jsonwebtoken/impl/JwtMap.java
```diff
@@ -155,4 +155,16 @@ public void clear() {
...
+        return map.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
```
lhazlewood: Same

lhazlewood:
Thanks for the PR!

Issue #125, opened 2016-05-24
JWS ECDSA signatures should not be ASN.1 DER formatted


martinvw:
After some security auditing it appeared that when using JJWT with elliptic curve, only a part of the signature is taken into account when validating. Than means than one can append any value to it, it could be the letter 'a' or even an XSS string which could reflect back...

I wrote a test to explain the problem, I can create a pull-request for it. But note that I only added the test, I did not yet resolve the problem.

The test can be found here: https://github.com/martinvw/jjwt/blob/master/src/test/groovy/io/jsonwebtoken/SignaturePostfix.groovy

lhazlewood:
I found a way where we can test this accurately using RFC 7520 test vectors:

https://tools.ietf.org/html/rfc7520#section-4.3

The trick here is that ECDSA will automatically generate random bytes during signature calculation (without us controlling that) so we have to 'fake' it to ensure the random number generator returns the same data in the test vectors.

The `FixedSecureRandom` technique outlined in [this SO question](http://stackoverflow.com/questions/19904548/calculating-ecdsa-signature-in-java-as-per-an-rfc-test-vector?rq=1) looks like it could be the solution to that problem.

lhazlewood:
After researching this more, I'm fairly confident this is **not** a security issue - just a compatibility one.  IIUC the ECDSA signature verification algorithm will only use the first N bytes of the specified signature and ignore any remaining.  So appending data to an ECDSA signature before validation isn't a valid test.

What *does* need to be done however is to ensure JWS-compatible byte array encoding for ECDSA signatures.

Currently JJWT returns the signature bytes computed directly by the Java JCA implementation, which according to JCA documentation are ASN.1 DER bytes.  See [Algorithm Specifications](http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#algspec) --> **The Digital Signature Algorithm** table --> **Signature Algorithm** row

JWS ECDSA signatures however need to be formatted instead as the EC point R and S unsigned integers converted to byte arrays, concatenated.  From https://tools.ietf.org/html/rfc7515#page-45:

> The result of the digital signature is the Elliptic Curve (EC) point (R, S), where R and S are unsigned integers ... the R and S values ... [are] octet sequences representing big-endian integers ... The JWS Signature is the value R || S

In other words (pseudo code):

```
byte[] signature = R.toUnsignedInteger().toByteArray() + S.toUnsignedInteger().toByteArray()
```

[...]

PR #137, opened 2016-06-27
Fixed ECDSA Signing and verification

martintreurnicht:
Brought up to spec as per https://tools.ietf.org/html/rfc7515#page-45

Fixes issue https://github.com/jwtk/jjwt/issues/136 and https://github.com/jwtk/jjwt/issues/125


pom.xml
```diff
@@ -307,12 +307,13 @@
                                 <lineRate>100</lineRate>
                                 <branchRate>90</branchRate>
                             </regex>
+                            <regex>
+                                <pattern>io.jsonwebtoken.impl.crypto.ECDSA</pattern>
```
lhazlewood: 100% coverage is required for all new code - please add more tests (the exclusions above are for old/legacy code that isn't yet fully up-to-date)

src/main/java/io/jsonwebtoken/impl/crypto/ECDSA.java
```diff
@@ -0,0 +1,177 @@
...
+
+
+import io.jsonwebtoken.SignatureAlgorithm;
+import io.jsonwebtoken.lang.JOSEException;
+
+
+/**
+ * Elliptic Curve Digital Signature Algorithm (ECDSA) functions and utilities.
+ */
+public class ECDSA {
```
lhazlewood: We only accept 'utils' classes for things the JDK makes final (e.g. StringUtils).  This logic should probably go in the EllipticCurveProvider class?

src/main/java/io/jsonwebtoken/lang/JOSEException.java
```diff
@@ -0,0 +1,22 @@
+package io.jsonwebtoken.lang;
+
+
+/**
+ * Javascript Object Signing and Encryption (JOSE) exception.
+ */
+public class JOSEException extends Exception {
```
lhazlewood: We have a `JwtException` already.

src/main/java/io/jsonwebtoken/lang/JOSEException.java
```diff
@@ -0,0 +1,22 @@
+package io.jsonwebtoken.lang;
```
lhazlewood: I apologize that this isn't documented anywhere, but the 'lang' package is essentially only for classes that augment/enhance the Java language itself (i.e. things that would have been nice to have in the language).

lhazlewood:
I gave a look at the PR - thanks for taking the time to help make improvements!

I made some comments/suggestions - please refactor those (or ask follow-up) questions so we can accept the PR.

Finally, we need to make these changes backwards compatible so it does not break existing code/JWTs.

We basically need logic like:  "is it the RFC standard format?  Ok, verify as expected?  Or was it the old/legacy format?  If so, verify based on the old format"

martintreurnicht:
@lhazlewood I've made all the changes as per your request, let me know if there is anything else

src/main/java/io/jsonwebtoken/impl/crypto/EllipticCurveSignatureValidator.java
```diff
@@ -40,7 +40,10 @@ public boolean isValid(byte[] data, byte[] signature) {
         Signature sig = createSignatureInstance();
         PublicKey publicKey = (PublicKey) key;
         try {
-            return doVerify(sig, publicKey, data, signature);
+            int expectedSize = getSignatureByteArrayLength(alg);
+            //if the expected size is not valid for JOSE, fall back to ASN.1 DER signature
```
lhazlewood: Awesome - thanks for this backwards compatibility fix!  Can you please update the comment to be clear that this is for backwards compatible reasons only and not something that is normally part of the JWS specification?  We need to be clear about why we're doing this in case we forget in the future.

lhazlewood:
@martintreurnicht looks good!  Just one minor request around a comment explaining backwards compatibility and I think we can merge.

martintreurnicht:
@lhazlewood I've tried to be as descriptive as possible around the backwards compatibility code, let me know if i need to add anything else

lhazlewood:
@martintreurnicht looks great - thank you!

Issue #149, opened 2016-07-19
ECDSA curve misnamed in README


mike9005:
In the [Supported Features section](https://github.com/jwtk/jjwt/blob/master/README.md#specification-compliant) of the README, the NIST `P-521` curve is misnamed as `P-512` in the `ES512` JWS algorithm description.

Reference: the [RFC on ECC in TLS](https://tools.ietf.org/html/rfc4492#appendix-A).

lhazlewood:
Can you please submit a pull request?

mike9005:
I would but don't think I have access rights?

lhazlewood:
You fork the repo into your own github account, create a branch there, and then submit a pull request from your fork/branch to this repo.  Here's a helpful link:  https://help.github.com/articles/using-pull-requests/

HTH!

PR #150, opened 2016-07-21
Fix ES512 description typo in README

mike9005:
Resolves issue #149: the NIST `P-521` curve is misnamed as `P-512` in the `ES512` JWS algorithm description in the README.

Reference: the [RFC on ECC in TLS](https://tools.ietf.org/html/rfc4492#appendix-A).


lhazlewood:
Thank you!

PR #158, opened 2016-08-31
improve jwt parser memory allocation

benbenw:
re-use buffer instead of creating new ones
avoid creating unneeded buffers in the Strings util methods
Stop continuously copying array with StringBuilder#deleteCharAt
work directly on StringBuilder instead of creating a temporary String

test added to cover the modified methods


lhazlewood:
very nice!

PR #172, opened 2016-09-20
Implement type conversions of integral claim values

### LINKED ISSUE #142: Depending on the number value of custom claim, DefaultClaims#get() returns Integer or Long
Simple test case:
```groovy
  String claimName = "long"
  Long claimVal = new Long(5)
  String compact = Jwts.builder().setClaims([(claimName): claimVal]).compact();
  Claims claims = Jwts.parser().parse(compact).body as Claims
  // `claims.get` throws `RequiredTypeException`
  assert claims.get(claimName, Long) == claimVal
```
Here, Jackson finds that a `5` is small enough to fit in the `Integer` and so unmarshalls into it, instead of expected `Long`.

What makes matters worse is that when a value is large enough (`≥ 2^31`), Jackson will switch to `Long` and so code that naively always `get()`-s `Integer` will stop working:
```groovy
  String claimName = "long"
  Long claimVal = new Long((Integer.MAX_VALUE as long) + 5)
  String compact = Jwts.builder().setClaims([(claimName): claimVal]).compact();
  Claims claims = Jwts.parser().parse(compact).body as Claims
  // ok
  assert claims.get(claimName, Long) == claimVal
  // throws
  assert claims.get(claimName, Integer) != null
```

I was going to submit a pull request with addition of simple `if` in the `DefaultClaims#get`, but than thought about all those similar use cases with `Byte`-s, `Short`-s, `Float`-s and `Double`-s, started implementing them, got a bit overwhelmed and decided to ask here if it's even worth going into such deeps or an `if` would be enough. What do you think?

sainaen:
Jackson chooses the target type for JSON numbers based on their value,
while deserializing without correct typing information present.
This leads to a confusing behavior:

    String token = Jwts.builder()
        .claim("byte", (byte) 42)
        .claim("short", (short) 42)
        .claim("int", 42)
        .claim("long_small", (long) 42)
        .claim("long_big", ((long) Integer.MAX_VALUE) + 42)
        .compact();
    Claims claims = (Claims) Jwts.parser().parse(token).getBody();
    claims.get("int", Integer.class); // => 42
    claims.get("long_big", Long.class); // => ((long) Integer.MAX_VALUE) + 42
    claims.get("long_small", Long.class); // throws RequiredTypeException: required=Long, found=Integer
    claims.get("short", Short.class); // throws RequiredTypeException: required=Short, found=Integer
    claims.get("byte", Byte.class); // throws RequiredTypeException: required=Byte, found=Integer

With this commit, `DefaultClaims.getClaim(String, Class<T>)` will
correctly handle cases when required type is `Long`, `Integer`, `Short`
or `Byte`: check that value fits in the required type and cast to it.

    // ... setup is the same as above
    claims.get("int", Integer.class); // => 42
    claims.get("long_big", Long.class); // => ((long) Integer.MAX_VALUE) + 42
    claims.get("long_small", Long.class); // => (long) 42
    claims.get("short", Short.class); // => (short) 42
    claims.get("byte", Byte.class); // => (byte) 42

Fixes #142.

-----
Note: I decided against converting values to String and calling static `valueOf()`, because that would complicate and slow down everything without obvious (to me) gains. Please feel free to comment here if you see cases that would work with such approach and don't work with current.
Also, this is my first commit, so any suggestions regarding style are welcomed. :)

### DIFF
--- a/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
+++ b/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
@@ -120,8 +120,23 @@ public class DefaultClaims extends JwtMap implements Claims {
             value = getDate(claimName);
         }
 
+        return castClaimValue(value, requiredType);
+    }
+
+    private <T> T castClaimValue(Object value, Class<T> requiredType) {
         if (requiredType == Date.class && value instanceof Long) {
             value = new Date((Long)value);
+        }
+
+        if (value instanceof Integer) {
+            int intValue = (Integer) value;
+            if (requiredType == Long.class) {
+                value = (long) intValue;
+            } else if (requiredType == Short.class && Short.MIN_VALUE <= intValue && intValue <= Short.MAX_VALUE) {
+                value = (short) intValue;
+            } else if (requiredType == Byte.class && Byte.MIN_VALUE <= intValue && intValue <= Byte.MAX_VALUE) {
+                value = (byte) intValue;
+            }
         }
 
         if (!requiredType.isInstance(value)) {
--- a/src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
+++ b/src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
@@ -716,6 +716,37 @@ class JwtParserTest {
         }
     }
 
+    @Test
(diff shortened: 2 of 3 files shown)


lhazlewood:
Awesome PR! Thank you!

PR #175, opened 2016-09-30
Add ignore validation options and static methods for validation of claims independent to token parsing

selimok:
This pull request contains the features requested in #80 and additionally code refactoring to reuse claim validation codes in static validation methods. 

With these changes we can: 
- Parse token by ignoring expiration, notbefore claim values or signature problems.
- Do these validations explicitly and separately by calling static validation methods. 

Use case: As part of our token expiration strategy, we create tokens with short life spans and re-create new tokens if needed. To do this in more convenient way, we need aforementioned ignore methods while parsing. This way we can do additional security checks like XSRF token validation or check JWT integrity by validating signature seperately, even if the JWT token expired. 

P.S.: I must add new unit test as Java code, since I have no idea how to write code in groovy. Additionally I made some code refactoring and splitted huge methods into smaller. I hope the names of new methods are appropriate to existing (if any) naming conventions. 

Thanks.


lhazlewood:
Please write the tests in Groovy - we have standardized on it for testing and it's even easier than Java.  We won't accept a PR with Java tests or without 100% coverage.

lhazlewood:
Actually, we can convert the Java to Groovy - I don't want to discourage your PR.  If you can though, please try!

selimok:
Ok, i will give a try. Thanks. And sorry for the test coverage :(

lhazlewood:
No worries!  We're happy to have the help!

selimok:
Do you have an idea how can fix that https://coveralls.io/jobs/18754160/source_files/1580589642#L15 ?

Thanks.

src/main/java/io/jsonwebtoken/JwtParser.java
```diff
@@ -261,8 +261,37 @@
...
+
+    /**
+     * Do not validate the JWS digital signature.
+     * <p>
+     * <p>Note that this circumvents security features of JWT and JWS.</p>
+     *
+     * @return the parser for method chaining.
+     * @see SignatureException
+     */
+    JwtParser ignoreSignature();
```
lhazlewood: I'm not sure we want to allow this - it is not good for a spec-compliant parser to allow avoiding mandated security rules.  It's probably better to catch the exception and then ignore it if you want to - otherwise it is not obvious in code that you're ignoring security checks.

src/main/java/io/jsonwebtoken/JwtParts.java
```diff
@@ -0,0 +1,72 @@
+package io.jsonwebtoken;
+
+/**
+ * Container class which contains encoded header, payload and signature parts of
+ * the token.
+ *
+ * @since 0.8
+ */
+public class JwtParts {
```
lhazlewood: This is not entirely correct.  A JWT can be a JWS or JWE, and they have different numbers of 'parts' (Unsecured JWT = 2 parts, JWS = 3, JWE = 5).  Whatever implementation we'd have around that needs to account for any/all 3 of these scenarios, and probably typed accordingly, e.g. `JwsParts extends JwtParts` and `JweParts extends JwtParts`.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -198,229 +202,260 @@ public boolean isSigned(String jwt) {
...
+        }
+	}
+
+	private void validateClaims(Header header, Claims claims) {
+		if (claims != null) {
+
+			if (!this.ignoreExpiry) {
+	            //https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-30#section-4.1.4
+	            //token MUST NOT be accepted on or after any specified exp time:
+	            ValidationHelper.validateExpiration(header, claims, this.allowedClockSkewMillis, this.clock);
```
lhazlewood: No static/helper methods in the implementation please.  Such methods are not object-oriented - they are more procedural and harder to test.

src/main/java/io/jsonwebtoken/impl/ValidationHelper.java
```diff
@@ -0,0 +1,193 @@
...
+import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.Clock;
+import io.jsonwebtoken.ExpiredJwtException;
+import io.jsonwebtoken.Header;
+import io.jsonwebtoken.PrematureJwtException;
+
+/**
+ * This class contains static methods for validation of JWT claims.
+ */
+public class ValidationHelper {
```
lhazlewood: Please no 'Helper' or 'Utils' or such classes for anything other than things that augment the Java language itself (e.g. Strings.whatever, Objects.whatever, Assert.whatever are fine because they augment the Java language and those things are often not overridable - such as String being final) - these typically often represent procedural (non OO) logic.

If anything, this would be a `Validator` or something similarly named.  Additionally, anything that can be configured/overridable should be an interface and a default implementation provided so that people can plug in different implementations or override them as desired, e.g. `Validator` and `DefaultValidator` (assuming it is correct to have such a component).

lhazlewood: Upon further review of this class, it appears the most OO/pluggable way for this it to have a generic `Validator` interface and you could have multiple implementations, e.g. `ExpirationValidator` `PrematureValidator`, etc.  For example:

```java
public interface Validator<T> {
    public void validate(T value);
}
```

And the type used by the parser would be `Validator<Claims>`

src/main/java/io/jsonwebtoken/JwtParts.java
lhazlewood: Because JWE is being implemented in the `jwe` branch, you might not need to worry about jwe parts yet, but whatever implementation we create should be extensible enough to support that upcoming feature/release.

lhazlewood:
I added a code review - the use of the static class is why you're getting code coverage failures.  HTH!

src/main/java/io/jsonwebtoken/JwtParser.java
selimok: I think calling this method by parsing the token is obvious enough. 
The way suggest (catching exception and ignoring it) is not really a good idea. Lets take a look in following scenario:

- We want to ignore expiry but do all other validations (nbf, signature, etc.)
- If we catch ExpiredJwtException we can surely say the token is expired, but other validations will not be executed, because with thrown exception, execution of parsing will be stopped immediately. So we can not ensure if nbf claim is valid or not. Currently signature validation comes first, but in the future the implementor may decide to change the order of validations. In this case we can also not sure if token signature valid at all. 

So we need a mechanism to say explicitly, to that validations but this not. If you see this use case not a common one. We can stop the discussion here :)

selimok: Ok we can remove this method. Security comes first. :)

src/main/java/io/jsonwebtoken/impl/ValidationHelper.java
selimok: Ok i will made some changes and remove that static methods. Thanks for advice.

selimok: The problem with Validator interface, that we cannot put additional method parameters into the validate method. For example for checking expires claim we need to optionally provide allowedClockSkewMillis or clock. Do you have any suggestion for this?

selimok:
Hi @lhazlewood, can you please check my changes again. Thx. :)

src/main/java/io/jsonwebtoken/JwtParser.java
```diff
@@ -338,6 +358,60 @@
...
+
+	/**
+	 * Checks if the specified compact serialized JWT string is expired based on the builder's current configuration state. 
+     * <p>
+     * <p>Note that you need this method probably only then, if the parser is configured to by pass 'expiration time' ({@code exp}) claim validation (by calling builder's {@link #ignoreExpiry()} method).</p>
+     *
+     * @param jwt the compact serialized JWT to check
+     * @return {@code true} if the specified JWT compact string is expired, {@code false} otherwise.
+	 */
+	boolean isExpired(String jwt);
```
selimok: Do you mean isPremature or isExpired or both? In our use case we need really first ignore the expiration validation and then separately check if the token expired or not. I don't really know, if there are other cases for isPremature out there. From my point of view, we can kick all ignoreNotBefore, isPremature and validateNotBefore methods entirely. But why not let them there, may be some one need them for a specific use case.

src/main/java/io/jsonwebtoken/JwtParser.java
```diff
@@ -338,6 +358,60 @@
...
+	/**
+	 * Validates 'expiration time' ({@code exp}) claim of the specified compact serialized JWT string based on the builder's current configuration state. 
+     * <p>
+     * <p>Note that you need this validation method probably only then, if the parser is configured to by pass 'expiration time' ({@code exp}) claim validation (by calling builder's {@link #ignoreExpiry()} method).</p>
+     *
+     * @param jwt the compact serialized JWT to check
+     * 
+     * @throws ExpiredJwtException if the validation of 'expiration time' ({@code exp}) claim is failed.
+	 */
+	void validateExpiration(String jwt);
```
selimok: isExpired method is just a wrapper on validateExpiration. I explained already above in my first comment, why we need isExpired method. So in my case again we can let validateExpiration method private. But _again_ why to remove this method? Both validateExpiration and validateNotBefore methods was in the code already, just there was a part of the huge parse method. I only moved that parts of the code into the smaller methods and made them accessible. So the decision is up to you.

src/main/java/io/jsonwebtoken/JwsParts.java
```diff
@@ -0,0 +1,32 @@
+package io.jsonwebtoken;
```
selimok: Right. I will change this.

selimok:
Any decisions?

selimok:
Are there anything else that I can do for a PR acceptance?

selimok:
If my PR less likely to be accepted, I would open a new feature request for setting clock skew seperately for notBefore and expiry claims. So that I can at least explicitly set a great amount of clock skew to simulate ignore expiry behaviour. Currently if I set the clock skew value, it affects both nb and exp claim validation. 

Thanks.

lhazlewood:
Hi @selimok - sorry - we just haven't had time to review this due to conferences (Gartner, re:Invent) and holiday/end-of-quarter deadlines.  We'll review this as soon as we can!

selimok:
Hi @lhazlewood, no problem. Sorry for my impatience.
Thanks in advance for your effort.

lhazlewood:
@selimok No worries! I'm sorry for the delay!  We really appreciate your patience, and we're grateful for the PR!

torben-amann:
Hey, what's the status of this pr? We are also trying to accomplish a validation without expiration date even though it's contained in the jwt.

pvegh:
Wow @selimok delivers a PR and all the requested changes but 6 months later there is still no feedback from the devs. 6 months!
This is a great example how not to handle an open source project.
If you have no time to support it anymore please add it to the README.MD so that others can decide if they really want to integrate this or look for another solution.

emeraquino:
Any update for the future of this PR?

hudson155:
Any update on this?

SeanLMcCullough:
Bump, this feature is important for a project I am working on

jonalbertini:
Years after the PR, all the quick fix @selimok done. still not merged... now I bet too much conflicts to merge... contribute they said...

lhazlewood:
@jonalbertini we're taking a different route with a proper `Validator` interface/builder with chaining methods and such for a future release.  This one slipped through the cracks because it was too specialized and not flexible enough for custom validation.  We're going with a different approach.

lhazlewood:
Closing just as a matter of housekeeping in favor of the work that will be done via #474.  If you feel otherwise, please let us know and we can discuss.

PR #176, opened 2016-10-04
Adding support for JCA Name along with Name on SignatureAlgorithm

sandeepkaul:
Adding support for Headers of the type:

{
  "typ": "JWT",
  "alg": "HmacSHA256"
}

### DIFF
--- a/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
+++ b/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
@@ -264,6 +264,9 @@ public enum SignatureAlgorithm {
             if (alg.getValue().equalsIgnoreCase(value)) {
                 return alg;
             }
+            if (alg.getJcaName().equalsIgnoreCase(value)) {
+              return alg;
+          }
         }
 
         throw new SignatureException("Unsupported signature algorithm '" + value + "'");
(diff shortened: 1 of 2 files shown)


lhazlewood:
It is unclear to me why this pull request was created - can you please clarify for us why it was opened?

JJWT's primary goals are spec compliance and ease of use.  The proposed change is not JWT specification compliant, and does not add any functionality that is not already possible using existing JWT-approved names.

sandeepkaul:
Hi,

We are trying to generate a token via https://github.com/pauldijou/jwt-scala, and we are using the jjwt library to validate the token on the other end. The jwt-scala library is generating the header, with alg field as HmacSHA256, which is same as HS256 at our end.

Now our library does not handle that Algorithm. But it is the same thing as HS256. 

Please suggest if there is a better way of doing this?

lhazlewood:
I think you should open a bug for the jwt-scala project - it appears that they are not specification compliant.  The JWT specification _requires_ that a JWT library must correctly support the `HS256` alg token, and if they claim to be a JWT library - as their name implies - they should fix their code to do the right thing.  Im not particularly excited about supporting a special behavior that no one else has asked us about , especially because another library is doing something incorrectly.  Does this make sense?

sandeepkaul:
Sure. makes sense.

Issue #189, opened 2016-11-14
Allow Exceptions to be throws in SigningKeyResolver interface


prateeknischal:
There is a use case that requires me to throw custom exceptions during resolving a key to validate the token. Suppose the token contains the client information which is required to get the key assigned to him. Now if an invalid user_id is passed in the token then It would be a client error. If the DB which has the key fails then it would not be able to retrieve keys even if the user_id is authentic which is a server side error.

In the current situation I am not able to differentiate between a client side error and a server side error because of being able to return a correct key or NULL.

If there is a "throws Exception" condition then I could use proper exceptions that are generated to categorise the errors.

gposcidonio:
So a short term solution may be to use unchecked exceptions which is less than ideal.

That having been said, it may be the case that you should do some validation on the token before passing it to the `SigningKeyResolver` which will ensure that any time you return `null`, you know that it's a database issue. This makes sense from a high level as it's not the SigningKeyResolver's job to do validation on the token.

prateeknischal:
It is kind of critical so using unchecked exception would not be accepted. All the information regarding the signing key is in the token itself, and cannot be guaranteed to be correct. Is there a way to get the header and claims parsed without passing the SigningKeyResolver or the actual signingKey (the latter would required increased and redundant information to be sent through a network).

#67 shows a way to get the claims without sign verification. But that would be extra computation, parsing the claims and then get the SigningKey, then again pass it to the claims to verify (which would again parse the token i guess).

SingingKeyResolver's job is to get the signing key which may live in a DB under hardware security, which require proper handling of errors and returning appropriate error codes for the client to be verbose enough. My intention is not to do validation on the token but to extract the information required to get the corresponding signing key.

lhazlewood:
The JJWT team feels strongly about using unchecked exceptions.  Even if the signature was changed to `throws Exception`, that exception would be wrapped and re-thrown as a `JwtException` (or some such subclass) because `parse` methods throw `JwtException`, not a checked exception.

Based on this, the correct solution is to throw a runtime exception and catch it, or catch JJWT's runtime exception, and expect `getCause()` if you desire.

PR #191, opened 2016-11-18
added support for throwing custom exceptions in SigningKeyResolver

prateeknischal:
Added support for throwing custom exceptions in SigningKeyResolver to get more granular errors while retrieving the key for sign verification.

Issue : #189

### DIFF
--- a/src/main/java/io/jsonwebtoken/JwtParser.java
+++ b/src/main/java/io/jsonwebtoken/JwtParser.java
@@ -287,7 +287,7 @@ public interface JwtParser {
      * @see #parsePlaintextJws(String)
      * @see #parseClaimsJws(String)
      */
-    Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException, SignatureException, IllegalArgumentException;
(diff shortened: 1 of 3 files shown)


gposcidonio:
This solution seems a little clunky to me. Since all of these methods now throw an arbitrary exception, you'll either have to catch Exception everywhere:

```
try {
...
} catch (Exception e) {
...
}
```

or you're going to have to declare that all your methods interacting with one of the methods you modified throw an Exception:

```
public void myMethodWhichUsesSigningKeyResolver() throws Exception { ... }
```

Both of these things seem to pollute any self-respecting codebase with unnecessary boilerplate. Furthermore, if you decided to start catching the exceptions (i.e. the first solution I mentioned above) then you'll start catching unchecked exceptions which will definitely obscure some behavior in your codebase.

Personally, I think it's not the responsibility of JJWT to resolve #189.

prateeknischal:
SigningKeyResolver is only used once in parsing the token, so I do not see a whole lot of exception catching, it should be just one function for one flow. Anyways when I think of it, seems that JWT should not handle my custom exceptions, So then I need to find a work around then.

PR #194, opened 2016-12-05
Added support for pluggable custom Crypto Implementation

prateeknischal:
This commit allows a user to plug their own crypto Implementation to the jjwt library and use it instead of the default provided Crypto layer. This may be required becuase the jjwt expects the user to get a private key to sign the token. But there might be cases that one cannot get the clear private key out, as in some applications (e.g. PCI compliant applications).

These applications use a HSM (https://en.wikipedia.org/wiki/Hardware_security_module) for all the encryption and decryption and key management. So the application never has access to clear keys, but only a handle. As far as I know there is no mechanism in the current build to have tokens instead of actual clear Key for crypto functions. Also the library relies completely on the default crypto layer provided. there must a mechanism to set custom crypto implementations that is flexible enough to accept tokens / blobs of actual keys and delegate the crypto functions to the centralised module (eg HSM). 

The above commit has such features. 

It adds 2 new data members CryptoProvider and Map<String, Object>, CryptoProvider being the flexible interface that provides sign() and verify() functions for the token and the map being the extra parameters that are required to do so (The provider might require some sort of identifier or a different algorithm / parameters to sign and verify).

Now one can pass the cryptoProvider using setCryptoProvider() and the config as setConfigParams(), also added in the DefaultJwtParser/Builder interfaces.

The user defined Crypto function need to implement the added CryptoProvider interface and pass an instance of that to the Jwts builder / parser. The parsing and error checking of the config map is left to the implementation of CryptoProvider as it is completely implementation / use case specific.

The original specs of the project are not violated in any way. Only if the cryptoProvider is set, the flow moves to the custom implementation, otherwise it will continue the default implementation, So completely backwards compatible (all tests are passing).

### DIFF
--- a/src/main/java/io/jsonwebtoken/CryptoProvider.java
+++ b/src/main/java/io/jsonwebtoken/CryptoProvider.java
@@ -13,7 +13,7 @@ public interface CryptoProvider {
 	 * 
 	 * Parsing of the config map is left for the implementation
 	 */
-	public String sign(String plain, Map<String, Object> config);
+	public String sign(String plain, Map<String, Object> config) throws SignatureException;
 	
 	/**
 	 * 
@@ -24,6 +24,6 @@ public interface CryptoProvider {
 	 * 		an instance for using jjwt. 
 	 * @return Boolean value identifying if the sign is valid or not
 	 */
-	public Boolean verify(String plain, String sign, Map<String, Object> config);
+	public Boolean verify(String plain, String sign, Map<String, Object> config) throws SignatureException;
 	
 }
--- a/src/main/java/io/jsonwebtoken/CryptoProvider.java
+++ b/src/main/java/io/jsonwebtoken/CryptoProvider.java
@@ -1,0 +1,29 @@
+package io.jsonwebtoken;
+
+import java.util.Map;
+
+public interface CryptoProvider {
+	/**
+	 * 
+	 * @param plain the string that needs to be signed
+	 * @param config configuration hashmap that requires exclusive information required for sign
+	 * 		This config map can be implementation specific as the user will have to implement the CryptoProvider and provide
+	 * 		an instance for using jjwt.
+	 * @return String the signed value as a String
+	 * 
+	 * Parsing of the config map is left for the implementation
+	 */
+	public String sign(String plain, Map<String, Object> config);
+	
+	/**
+	 * 
+	 * @param plain the string that has been signed
+	 * @param sign the sign that is to be verified for the given plain text
+	 * @param config configuration hashmap that requires exclusive information required for verify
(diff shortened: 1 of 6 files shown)


lhazlewood:
We're already working on such a change in the `jwe` branch where the `SignatureAlgorithm` enum becomes an interface so you can provide your own signature algorithms.

lhazlewood:
Also, FWIW, any config necessary for the SignatureAlgorithm implementation would be supported in the implementation class, not the `SignatureAlgorithm` interface itself.  A hint as to why is that all current algorithms do not need a config Map at all: that your algorithm might need one is a reflection that the very nature of the Map is implementation specific and therefore shouldn't be in the general-purpose interface.

Based on this, you would instantiate the SignatureAlgorithm implementation with the config as necessary and then invoke the builder with it.  For example:

```java
Map<String, Object> config = getMyConfig();
SignatureAlgorithm alg = new MySignatureAlgorithm(config);
Jwts.builder()
   // ... etc ...
   .signWith(alg, key)
   .compact();
```

This is the approach we'll be taking in the `jwe` branch.  HTH!

prateeknischal:
But the main concern is the ".signWith(alg, key)" part. The singer needs to have the clear private key to sign which is not recommended in PCI compliant applications rather not allowed. So all I have is a token and a few meta fields corresponding to the actual key. Since the HSM module does not have an API to give out the clear private key, I can only pass the encrypted token and meta fields.

The HSM has API for sign and verify but that cannot be embedded in the jjwt library itself. Which is why I need the CryptoProvider interface. The map part is not that important but the custom crypto provider is. 

I think that providing a config map allows users to pass all the necessary arguments to their Algo implementation. 
Also I believe createNewSigner(alg, key) creates a new instance of the signer which cannot be done in an HSM, we need a single instance because of all the init stuff it does. So passing a custom implementation of signer would be beneficial. 

The SignatureAlgorithm interface is a great plus. This is in a way similar to it but instead of createNewSigner() we can have setSigner() or the setCryptoProvider() which would be much more flexible.

Please review the code and the test case too for my use case. If there is some dev branch where this could be worth merging then I'll be happy to create a pull req there.

prateeknischal:
The the current design and code in parse/build would have to be changed, right?

lhazlewood:
@prateeknischal yes, once the `SignatureAlgorithm` change to an interface is complete, you'll be able to supply your own implementation as desired, it's just those won't be standard/RFC-approved (which shouldn't matter if no-one outside your organization needs to verify those signatures).

lhazlewood:
and yes, the current code in DefaultJwtParser#parse DefaultJwtBuilder#compact would have to change.  This is an implementation detail, so it shouldn't impact anyone's code.

Issue #196, opened 2016-12-19
Allow bulk adding of claims


roberterdin:
Can you introduce a method in the `JwtBuilder` that allows for bulk insertation of claims but does not overwrite previously set ones?

```
Jwts.builder()
     .setClaims(someClaims) 
    .setSubject(someSubject)
    .signWith(SignatureAlgorithm.HS256, key)
    .compact();
// --> all good
```

```
Jwts.builder()
    .setSubject(someSubject)
    .setClaims(someClaims) 
    .signWith(SignatureAlgorithm.HS256, key)
    .compact();
// --> subject gets overwritten :(
```

While it is easy to avoid the overwriting it just makes the code prone to errors when the order of the builder functions matters.

**Proposed solution.**
Change the `JwtBuilder` and the `Claims` interfaces and implementations such that they allow an addClaims functions like this one in the `DefaultJwtBuilder` implementation.
```
    @Override
    public JwtBuilder addClaims(Map<String, Object> claims) {
        this.claims.putAll(claims)
        return this;
    }
```

I'm happy to submit a PR.

Cheers

lhazlewood:
Thanks for the issue!  A PR would be appreciated!  Please add an `@since 0.7.0` JavaDoc tag to the method before submission.

petergphillips:
I've just been caught out by the `setClaims` method too, but for a different reason.  The `setClaims` method takes a `Map` and then modifies the passed in map by adding the other claims to the map.  This means that  a immutable map (e.g. `ImmutableMap` or `Collections.unmodifiableMap`) can't be used in the builder.  Modification of input parameters by an API is also not pleasant.

Would it be worth therefore deprecating the `setClaims` method altogether and moving to `addClaims` instead?

lhazlewood:
PR merged - thank you!

PR #200, opened 2017-01-10
Added addClaims function to JwtBuilder as described in Issue #196.

roberterdin:
This function does not overwrite existing claims (as compared to setClaims). See Issue #196


lhazlewood:
Thanks for the PR!

PR #203, opened 2017-01-24
make_objectmapper_static

wasifaleem:
Makes ObjectMapper static in DefaultJwtParser.
See: http://wiki.fasterxml.com/JacksonBestPracticesPerformance

### DIFF
--- a/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
+++ b/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
@@ -59,7 +59,7 @@ public class DefaultJwtParser implements JwtParser {
     private static final String ISO_8601_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
     private static final int MILLISECONDS_PER_SECOND = 1000;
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -59,7 +59,7 @@
     private static final String ISO_8601_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
     private static final int MILLISECONDS_PER_SECOND = 1000;
 
-    private ObjectMapper objectMapper = new ObjectMapper();
```
lhazlewood: I'm not sure we want to do this - it might be configurable in the future to configure a parser with its own objectmapper instance (e.g. for pretty printing or whatever)

kpronovost: I agree with @lhazlewood 
It'd be better to be able to set the ObjectMapper. That being said, it is currently not the case and I would say it would be better to be consistant throught the different classes.

If the static is added though, please rename to OBJECT_MAPPER

lhazlewood:
This has been resolved via #344

PR #209, opened 2017-02-15
Update README.md

arlemi:
Added missing Variable Declarator for the IncorrectClaimException catch that would raise an error in Eclipse.


lhazlewood:
Thank you for the PR!

PR #214, opened 2017-03-14
Added POJO deserialisation to get(Object, Class<T> requiredType)

cbudtz:
When you can do this:
```
String token = Jwts.builder()
				.claim("user", new User(1L,"test","testPass"))
				.signWith(SignatureAlgorithm.HS512, key)
				.compact();
```
shouldn't you be able to do this?:
```
User usr = Jwts.parser().
setSigningKey(key).
parseClaimsJws(tokenString).
getBody().
get("user", User.class);
```
There might be a good reason why not?

### DIFF
--- a/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
+++ b/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
@@ -128,9 +128,7 @@ public class DefaultClaims extends JwtMap implements Claims {
     private <T> T castClaimValue(Object value, Class<T> requiredType) {
         if (requiredType == Date.class && value instanceof Long) {
             value = new Date((Long)value);
-        }
-
-        if (value instanceof Integer) {
+        } else if (value instanceof Integer) {
             int intValue = (Integer) value;
             if (requiredType == Long.class) {
                 value = (long) intValue;
@@ -139,9 +137,7 @@ public class DefaultClaims extends JwtMap implements Claims {
             } else if (requiredType == Byte.class && Byte.MIN_VALUE <= intValue && intValue <= Byte.MAX_VALUE) {
                 value = (byte) intValue;
             }
-        }
-
-        if (!requiredType.isInstance(value)) {
+        } else if (!requiredType.isInstance(value)) {
         	try {
         		return new ObjectMapper().convertValue(value, requiredType);
         	} catch (IllegalArgumentException e) {
--- a/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
+++ b/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
@@ -21,6 +21,8 @@ import io.jsonwebtoken.RequiredTypeException;
 import java.util.Date;
 import java.util.Map;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 public class DefaultClaims extends JwtMap implements Claims {
 
     public DefaultClaims() {
@@ -140,7 +142,11 @@ public class DefaultClaims extends JwtMap implements Claims {
         }
 
         if (!requiredType.isInstance(value)) {
-            throw new RequiredTypeException("Expected value to be of type: " + requiredType + ", but was " + value.getClass());
+        	try {
+        		return new ObjectMapper().convertValue(value, requiredType);
+        	} catch (IllegalArgumentException e) {
+        		throw new RequiredTypeException("Expected value to be of type: " + requiredType + ", but was " + value.getClass());
+        	}
         }
 
         return requiredType.cast(value);
--- a/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
+++ b/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
@@ -25,126 +25,130 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 public class DefaultClaims extends JwtMap implements Claims {
 
-    public DefaultClaims() {
-        super();
-    }
+	public DefaultClaims() {
+		super();
+	}
 
-    public DefaultClaims(Map<String, Object> map) {
-        super(map);
-    }
+	public DefaultClaims(Map<String, Object> map) {
+		super(map);
+	}
 
-    @Override
-    public String getIssuer() {
-        return getString(ISSUER);
-    }
+	@Override
+	public String getIssuer() {
+		return getString(ISSUER);
+	}
 
-    @Override
-    public Claims setIssuer(String iss) {
-        setValue(ISSUER, iss);
-        return this;
-    }
+	@Override
+	public Claims setIssuer(String iss) {
+		setValue(ISSUER, iss);
+		return this;
+	}
 
-    @Override
-    public String getSubject() {
-        return getString(SUBJECT);
-    }
+	@Override
+	public String getSubject() {
+		return getString(SUBJECT);
+	}
 
-    @Override
-    public Claims setSubject(String sub) {
-        setValue(SUBJECT, sub);
-        return this;
-    }
+	@Override
+	public Claims setSubject(String sub) {
+		setValue(SUBJECT, sub);
+		return this;
+	}
 
-    @Override
-    public String getAudience() {
-        return getString(AUDIENCE);
-    }
+	@Override
+	public String getAudience() {
+		return getString(AUDIENCE);
+	}
 
-    @Override
-    public Claims setAudience(String aud) {
-        setValue(AUDIENCE, aud);
-        return this;
-    }
+	@Override
+	public Claims setAudience(String aud) {
+		setValue(AUDIENCE, aud);
+		return this;
+	}
 
-    @Override
-    public Date getExpiration() {
-        return get(Claims.EXPIRATION, Date.class);
-    }
+	@Override
+	public Date getExpiration() {
+		return get(Claims.EXPIRATION, Date.class);
+	}
 
-    @Override
-    public Claims setExpiration(Date exp) {
-        setDate(Claims.EXPIRATION, exp);
-        return this;
-    }
+	@Override
+	public Claims setExpiration(Date exp) {
+		setDate(Claims.EXPIRATION, exp);
+		return this;
+	}
 
-    @Override
-    public Date getNotBefore() {
-        return get(Claims.NOT_BEFORE, Date.class);
-    }
+	@Override
+	public Date getNotBefore() {
+		return get(Claims.NOT_BEFORE, Date.class);
+	}
 
-    @Override
-    public Claims setNotBefore(Date nbf) {
-        setDate(Claims.NOT_BEFORE, nbf);
-        return this;
-    }
+	@Override
+	public Claims setNotBefore(Date nbf) {
+		setDate(Claims.NOT_BEFORE, nbf);
+		return this;
+	}
 
-    @Override
-    public Date getIssuedAt() {
-        return get(Claims.ISSUED_AT, Date.class);
-    }
+	@Override
+	public Date getIssuedAt() {
+		return get(Claims.ISSUED_AT, Date.class);
+	}
 
-    @Override
-    public Claims setIssuedAt(Date iat) {
-        setDate(Claims.ISSUED_AT, iat);
-        return this;
-    }
+	@Override
+	public Claims setIssuedAt(Date iat) {
+		setDate(Claims.ISSUED_AT, iat);
+		return this;
+	}
 
-    @Override
-    public String getId() {
-        return getString(ID);
-    }
+	@Override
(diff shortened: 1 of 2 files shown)


lhazlewood:
jjwt does not currently support typed-object claims yet.  Does this PR support that functionality fully?

cbudtz:
Frankly - I don't have a full picture of how typed-objects behave in jjwt. This PR only adresses what seemed to me to be an inconsistency - Namely that you can pass a typed object as a claim, but don't retrieve it without mapping it yourself. I only added an attempt at mapping the object to the passed type.

The implementation already maps Dates, Longs, Ints and Shorts - I merely expanded the solution to try mapping any other types as well. 
UnitTest should be fixed first, though....

cbudtz:
Now fixed silent loss of precision and wrote unit test to rectify code coverage loss.

cbudtz:
Any decisions on my feature?

cbudtz:
Any decision on this feature?

cbudtz:
Any news on my pull request? It's been 7 months now...

lhazlewood:
@cbudtz I'm terribly sorry for the long delay - things were insanely busy for us then.  That said, things are in full swing again and we've gotten a lot of work done on JJWT recently, currently planning for the 0.10.0 release.  Now that #344 has been merged to master, and you can customize how the objectmapper works, is this PR still necessary?  If so, what needs to be done to complete it?

cbudtz:
Thanks for getting back to me.

#344 doesn't address the issue that my PR is adressing - I still miss the feature proposed in my PR - Which is an attempt at deserializing the Object to the required type. Jjwt already does this with Long, Short and Byte - so i think its a valid and useful feature. 

I Guess 
`value = new ObjectMapper().convertValue(value,requiredType);`

should be rewritten to use the new InstanceLocator, but frankly its a little obscure for me how that should be done - something along this line i guess??:
` InstanceLocator<Deserializer<Map<String,?>>> locator =
                        Classes.newInstance("io.jsonwebtoken.impl.io.RuntimeClasspathDeserializerLocator");
                Deserializer<Map<String, ?>> instance = locator.getInstance();`

If you think its a good feature, I'd like some help with conforming to #344 , but if I'm the only one wishing for this feature, it might not be worth the effort... I can always deserialize the claim by hand - though it feels a little weird.

lhazlewood:
> value = new ObjectMapper().convertValue(value,requiredType);

@cbudtz the problem with this is that the entire JSON payload (a String) has been transformed into a Java `Map` (which could contain any json value, including nested arrays and maps, etc, etc) _before_ the Claims are exposed to a user.

The `InstanceLocator` is only used if you don't specify a custom Deserializer (via `JwtParser#deserializeJsonWith`) and defaults to whatever JSON marshalling mechanism is available (e.g. Jackson or org.json).

This means you could do something like this:

```java
ObjectMapper om = getMyObjectMapper(); //implement me
Jwts.parser()
    .setDeserializer(new JacksonDeserializer(om))
    // ... etc
    .parsClaimsJws(jws);
```

The `JsonParser` currently requires a `Map<String,Object>` to be returned from the JSON `Deserializer` in all cases.  This means that a *nested* property within that map needs to be converted to a domain-specific Object, but not any others, and this needs to be known at _parsing time_, not at read time.

Could you implement `getMyObjectMapper()` such that it returns a custom ObjectMapper that knows to deserialize a standard map, but with a POJO deserializer for only a specific property within the map?  I don't know, I've never done that before.  Here's a quick StackOverflow answer that might help:

https://stackoverflow.com/questions/36721597/parsing-a-subset-of-json-in-java-using-jackson

That said, the current way people do this is to convert the value to a string before creating the JWT and then reverting that string to a java object after parsing.  For example:

```java
String serializedUser = objectMapper.writeValueAsString(user);
String jws = Jwts.builder()
    .claim("user", serializedUser)
    // ... etc...
   .compact();
```

And then after parsing:

```java
String serializedUser =  Jwts.parser()
    .setSigningKey(key)
    .parseClaimsJws(jws)
    .getBody().
    .get("user", String.class);

User user = objectMapper.readValue(serializedUser, User.class);
```

Based on that and how easy it is, does JJWT need to support anything more?  I think it'd be great to automate this if we can keep it simple, but providing a custom Deserializer may be the best way - I'm not sure.  Thoughts?

cbudtz:
Well I think that my original argument still stands. When you can
`.claim("user", new User(1L,"test","testPass"))`
and get automatic serialization,
then the reverse should be possible as well:
`.get("user", User.class);`
(Especially since Data,Long, Int, Short and Byte are automatically deserialized).

Since the serialization process depends on the objectmapper - which can just as easily reverse the process it seems like it would be an obvious feature to include. If you dont include automatic deserialization, maybe you should remove the option for automatic serialization typed claims to avoid inconsistency - meaning getting rid of 
`Jwts.builder()
				.claim(String, Object)`
and only preserve
`Jwts.builder()
				.claim(String, String)`


> This means that a nested property within that map needs to be converted to a domain-specific Object, but not any others, and this needs to be known at parsing time, not at read time.

Just to be sure - 'parsing time' being when Json is mapped to claims and 'read time' being when a claim is extracted? What is wrong with mapping a claim to a POJO on .get("claimName", ClaimClass.class))

I'm sorry if im missing the point, but its not clear to me why you cant have an attempt at deserializing a typed claim in DefaultClaims.get(String claimName, Class<T> requiredType) - Is it because of the new deserializer interface that only provides a deserialize(byte[]) ? I guess this would neccissitate expanding the interface with 

`public T convert(Object pojo)`

which seems like  a pretty big deal to implement my feature request...

In short - maybe you should just get rid of: 
`Jwts.builder()
				.claim(String, Object)`

Since that leads me to expect the reverse is possible as well...

lhazlewood:
> When you can `.claim("user", new User(1L,"test","testPass"))` and get automatic serialization

That's a side-effect of Jackson - the JSON serializer implementation - performing automatic POJO serialization - not something built-in to (or even considered in the design of) JJWT.  If you use the `OrgJsonSerializer` for example, you may see entirely different results.

> Just to be sure - 'parsing time' being when Json is mapped to claims and 'read time' being when a claim is extracted?

Correct.

> What is wrong with mapping a claim to a POJO on .get("claimName", ClaimClass.class))

Before you call `get` the JSON has _already_ been deserialized to a Map.  In other words, consider the following JSON:

```json
{
  "aud": "you",
  "exp": 1300819380,
  "user": {
    "name": "Joe",
    "address": {
        "street": "1234 Main Street",
        "city": "Anytown",
        "state": "CA",
        "postalCode": 94401
    }
  }
}
```

The `Deserializer` takes the raw UTF-8 bytes of this JSON String and returns a `Map<String,Object>` to coincide with the JWT specification definition of claims.  This implies that - unless you customize the JSON deserialization process - `claims.get("user")` will return a `Map<String,?>` (which itself has a `name` key and a `address` key, which in turn is also a `Map<String,?>`.

The reason `user` is a `Map` is because that's what Jackson unmarshalls by default - it has no way of knowing that that JSON value is supposed to correspond to a User object or not.  Unless you tell it as such by customizing the `ObjectMapper` instance.

So if, after parsing, `claims` is a `Map` (which might have nested Lists and Maps), you'd have to convert a Map to a User, not a JSON string to a User.

The only way to ensure that the `user` claim is *not* a Map after JSON parsing - but before you call `get` -  is to tell Jackson "Hey, when you see a "user" member in the JSON, I want you to convert it to a `User` instance instead of the default `Map` that you would ordinarily return".  You have to do that currently by providing a configured `ObjectMapper`.

Does this make sense?  I hope I'm conveying my thoughts properly!  Hopefully this helps and allows us to continue to brainstorm...

lhazlewood:
@cbudtz I'm not so sure that we should do option number 2: I know for sure people have asked for the ability to marshall POJOs as a claim, and I think some are doing it already.  If we deprecate that, they don't have a good option to do what they want.

After doing some research this morning, I'm not sure number 3 is necessary either.  You can customize the deserialization process (and therefore tell Jackson that you want unmarshalling of certain members) by doing this:

```java
ObjectMapper om = getMyObjectMapper(); //implement me
Jwts.parser()
    .deserializeJsonWith(new JacksonDeserializer(om))
    // ... etc
    .parsClaimsJws(jws);
```

I found a StackOverflow answer that (I think) shows how to do this in Jackson:

https://stackoverflow.com/a/27512093/407170

Alternatively, instead of customizing Jackson, the developer can just do the following 

```java
Map userMap = Jwts.parser()
    .setSigningKey(key)
    .parseClaimsJws(jws)
    .getBody().
    .get("user", Map.class);

User user = objectMapper.convertValue(userMap, User.class);
```

Based on these, I think just documenting how to do it would be sufficient rather than forcing a change in the API?  It seems to me like these approaches (especially the simplicity of the last one) gets us what we want, no?

lhazlewood:
P.S. I forgot to mention, I don't want option number 3 in the JJWT API itself because that *forces* JSON/Java unmarshaling on a JSON parser implementation, which is not trivial and can significantly bloat dependency size and isn't guaranteed to be supported by all parser implementations.  org.json (for example and the default API available to Android users) supports marshaling, but not unmarshaling.

lhazlewood:
Thinking about this fully, I think that if we implemented the StackOverflow answer's approach:

https://stackoverflow.com/a/27512093/407170

By allowing JacksonDeserializer to have a constructor that accepts a Map<String,Class> of fieldName-to-expectedType, then you could do this:

```java
User user = (User)
    Jwts.parser()
    .deserializeJsonWith(new JacksonDeserializer(Maps.of("user", User.class))
    .setSigningKey(key)
    .parseClaimsJws(jws)
    .getBody().
    .get("user");
```

The important part here is that this is behavior specific to Jackson and not something imposed on the wider-scope JJWT API that needs to be required of every parser implementation.

Thoughts?

cbudtz:
Seems like I have stirred up a hornets nest ;)

I think that 
`.deserializeJsonWith(new JacksonDeserializer(Maps.of("user", User.class))
.get("user")`

dosn't match the simple use (and automarshalling) of
`.claim("user", myUser)` 

Having to specify a specialized JacksonDeserializer to unmarshall a POJO, that was mashalled automagically doesn't _feel right_ .

I'm not so purist that I'll press the issue of deprecating the automatic marshalling, but just a little curious of why the wish for an automatic unmarshalling isn't present.

I think the pragmatic path of placing a _warning label_ on .claim(String, Object), telling that get(String, class) will only work for automatic unmarshilling of Date, Long, Integer, Byte and Short will suffice.

My initial pull request was motivated by an irritating debug session (which took far too long, mainly because I was a moron), that ended in the question: "Why on earth can't I unmarshall my claim with jjwt, when i can marshall it?".

This error message mislead me:
`
if (!requiredType.isInstance(value)) {
            throw new RequiredTypeException("Expected value to be of type: " + requiredType + ", but was " + value.getClass());
        } 
`

This is not very helpful in telling why it doesn't work - maybe it should be changed to

"Conversion failed. Expected value to be of type: " + requiredType + ", but was " + value.getClass() + ". JJWT will only convert Date, Long, Integer, Short and Byte automatically."

lhazlewood:
> just a little curious of why the wish for an automatic unmarshaling isn't present.

It's not that I don't wish for it :) It's because as I said above:

> org.json (for example and the default API available to Android users) supports marshaling, but not unmarshaling

So we can't really force every pluggable JSON implementation to support data binding/unmarshaling. 

Data binding requires an awful lot of logic and complexity in an implementation (and it's why [JSON-B](http://json-b.net/) is now finally a thing), so simpler parsers may not support it.  Also, JWT as a concept has no notion of data marshaling - it is purely a json processor-specific feature, which is why IMO that configuration should be relegated to the Deserializer implementation.

That said, your suggestion for improving the exception message makes total sense!  But I'd probably write it as the following to make it super clear for users that might encounter this:

`String msg = "Cannot convert existing claim value of type " + value.getClass() + " to desired type " + requiredType + ".  JJWT only converts simple Date, Long, Integer, Short and Byte types automatically.  Anything more complex is expected to be already converted to your desired type by the JSON Deserializer implementation.  You may specify a custom Deserializer for a JwtParser with the desired conversion configuration via the JwtParser#deserializeJsonWith method.";`

Based on that, plus:

- JavaDoc for the `get` method (dunno why it is missing)
- public documentation on marshaling/unmarshaling and how to do it w/ JJWT
- supporting `new JacksonDeserializer(Maps.of("user", User.class))`

Does that sum up the work to do to resolve this?

I just want to ensure we capture what we need to do so we can translate it to actionable work.  Thanks for your continued feedback!

cbudtz:
Seems perfectly reasonable, I'm closing the pull request.

joca-bt:
I believe `claims.get(<claim name>, class)` is very misleading. It gives the user the impression it can be any object, as per serialization, when it's quite not the case.

Issue #68, opened 2015-11-21
Android 6 RSAPrivateKey


KalinKanev:
Hello,

I understand that this may not be issue with the library but may is worth extending. Some changes were made by google in API 23 so the code below that worked on lower version is no longer valid.

    PrivateKey privateKey = KeyChain.getPrivateKey(context, "Alias");
    String jwtString = Jwts.builder().setHeaderParams(map)
                .setPayload(payload)
                .signWith(SignatureAlgorithm.RS256, privateKey)
                .compact();

Before API 23 KeyChain.getPrivateKey had a return type of OpenSSLRSAPrivateKey and post 23 it is AndroidKeyStoreRSAPrivateKey which does not expose the bytes of the key.

Could the library be enhanced to use the new Android way of signing as described in this post: 
https://code.google.com/p/android-developer-preview/issues/detail?id=2854

Thanks in advance,
Kalin

lhazlewood:
This is actually an easy fix I think.  Instead of asserting:

```java
(key instanceof RSAPrivateKey)
```

we can just assert:

```java
(key instanceof PrivateKey && key instanceof RSAKey)
```

This would assert the correct expectations and would work both on the standard JDK because the JDK's `RSAPrivateKey` interface extends both `PrivateKey` and `RSAKey`, and so does Android's RSA private key implementation.

lhazlewood:
This has been implemented and merged to master.  It will be in the next release.

KalinKanev:
Working like a charm. Thanks for fast reaction!

PR #224, opened 2017-05-25
Test for instanceof PrivateKey and instanceof ECKey when signing

woody77:
Similar to issue #68  (https://github.com/jwtk/jjwt/issues/68), EC keys on Android do not implement ECPrivateKey.  This changes the check in EllipticCurveSigner.java to use the same test as was used to solve issue #68 for RSA keys.

### DIFF
--- a/src/main/java/io/jsonwebtoken/impl/crypto/EllipticCurveSigner.java
+++ b/src/main/java/io/jsonwebtoken/impl/crypto/EllipticCurveSigner.java
@@ -19,6 +19,7 @@ import java.security.InvalidKeyException;
 import java.security.Key;
 import java.security.PrivateKey;
 import java.security.Signature;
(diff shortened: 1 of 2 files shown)


lhazlewood:
@woody77 should we have a test for this? I can't remember, but did the RSA change have an accompanying test?

woody77:
@lhazlewood - If you mean what I think, there is a test which checkx that the right exception is thrown when the wrong key type is used (testConstructorWithoutECPrivateKey).  This patch tweaks the test in the same way as the test for #68 was tweaked (the exception being thrown has a slightly different message in the same way as #68 changed the message, to add a space so that ECPrivateKey becomes EC PrivateKey).

lhazlewood:
Got it - sounds good, and thank you for the PR!

PR #240, opened 2017-07-25
add support for supplying multiple public keys, in order to gracefully support certificate rotation

thepeachbeetle:
We have a requirement for multiple public key to be valid for a period of time whilst performing a certificate rotation. A rotation across the whole global infrastructure is a non atomic operation that can potentially take some time to roll out. During this time, client requests may be initiated using either the current 'old' key, or a new key that is being rolled out, until such a time as the new key is fully rolled out.

It is therefore desirable for the service receiving the requests to be able to validate signatures generated with either of those keys.

This change exposes via the SigningKeyResolver interface, the ability to supply either a single key (using the existing signatures)  or a Collection of keys via the new methods added here. The remainder of the change should be transparent to the application, it is mainly plumbing to get the collection of keys to the various signature validator implementations and having them iterate over a list of possible keys untill one matches.

I appreciate that there would be further work needed to update the test suite, but i wanted to post this PR at this point first to get some feedback on this approach before going any further


thepeachbeetle:
superseded by PR #241

PR #242, opened 2017-07-26
Expose crypto implementation package in OSGI bundle headers

badfish69:
By default, maven-bundle-plugin excludes packages in private/impl namespaces from the Export-Package OSGI header.  This means that the classes in io.jsonwebtoken.impl.crypto aren't resolvable when deployed in an OSGI stack.  This change simply adds that package to the exports list.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -328,6 +328,11 @@
                             *
                         ]]>
                         </Import-Package>
+                        <Export-Package>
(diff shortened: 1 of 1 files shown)


lhazlewood:
We've had a package/module refactoring since this issue was created.  This is resolved via #502.

PR #246, opened 2017-08-12
Update README.md

Ledenel:
Add notice in examples.
The motivation to add this note is that I got an exception like "Signed Claims JWSs are not supported" while doing the quickstart. The reason is a typo `parseClaimsJwt` , which should be `parseClaimsJws`. Since they are both valid methods, IDE auto-completion doesn't help. It takes me a lot of time to find and fix it, and I think it's not friendly enough for a no-hint runtime exception that may happen in the quickstart guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,8 @@ Now let's verify the JWT (you should always discard JWTs that don't match an exp
 assert Jwts.parser().setSigningKey(key).parseClaimsJws(compactJws).getBody().getSubject().equals("Joe");
 ```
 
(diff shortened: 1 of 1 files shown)


lhazlewood:
Thanks for the PR!  Unfortunately this is more suitable for a documentation PR than a readme update.  We'll leave this issue open to ensure that we include this in proper documentation (rather than just Readme + Release notes).

lhazlewood:
Easier to just merge and deal with the docs later.  Thanks for the PR!

PR #268, opened 2017-10-10
Jwe java 9 update

dogeared:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## Release Notes
 
+### 0.9.0
+
+This is a minor release that includes changes to dependencies and plugins to allow for building jjwt with Java 9.
+Javadocs in a few classes were updated as well to support proper linting in both Java 8 and Java 9.
```
les-okta: Can we put in another sentence that says something like "JJWT currently works on Java 7, 8 and 9, but we will be targeting only 8 and above for 1.0 final" ?

PR #270, opened 2017-10-14
Refactored all occurrences of 'vector' to 'value' in the context of 'initialization vector'

dogeared:
per https://tools.ietf.org/html/rfc4949#page-9 as 'initialization vector' is DEPRECATED

### DIFF
--- a/src/main/java/io/jsonwebtoken/Jwe.java
+++ b/src/main/java/io/jsonwebtoken/Jwe.java
@@ -2,7 +2,7 @@ package io.jsonwebtoken;
 
 public interface Jwe<B> extends Jwt<JweHeader,B> {
 
-    byte[] getInitializationVector();
+    byte[] getInitializationValue();
 
     byte[] getAadTag();
 }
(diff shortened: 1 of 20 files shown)


lhazlewood:
I don't know how I feel about this:

1.  Any backwards incompatible change like this should be saved for the 1.0 release
2. Every cryptographic resource I've ever referenced calls this an Initialization Vector, and anyone who knows what this is or why/how it's used could be thrown off when seeing IntializationValue.  It's just 'weird' and confusing IMO.  I certainly was confused at first :)
3. We should call this whatever the JWT RFC calls it, regardless of whatever RFC4949 says.  That is, the JWT RFCs are our source of truth, and we should only look to other RFCs for things not already covered in the JWT RFCs.

dogeared:
I think your #3 is the most compelling argument for keeping it as it. Still, I am torn too. RFC4949 is pretty explicit:

```
$ initialization vector
      (D) /cryptography/ Synonym for "initialization value".

      Deprecated Term: To avoid international misunderstanding, IDOCs
      SHOULD NOT use this term in the context of cryptography because
      most dictionary definitions of "vector" includes a concept of
      direction or magnitude, which are irrelevant to cryptographic use.
```

The odd thing, is that RFC4949 is dated 2007 and RFC7516 is 2015.

I agree with the "source of truth" statement you made above. However, RFC7516 explicitly calls out RFC4949: https://tools.ietf.org/html/rfc7516#section-2

If you feel strongly about keeping it as-is, I'm game. But, I think it's worth having the spec clarified - either RFC4949 or RFC7516 should change, IMHO

lhazlewood:
If possible, I'd like to close this out since the JWT RFC calls this an Initialization Vector.  If you've no objections, could you please close it?

lhazlewood:
Closing this due to inactivity.  Please re-open if you'd like to continue discussion.

Issue #234, opened 2017-07-04
dependency-check-maven/1.4.5 plugin reporting a false CVE against jjwt/0.7.0


mkempster:
Running the org.owasp/dependency-check-maven plugin reports jjwt/0.7.0 has this (clearly wrong) CPE

cpe:/a:sonatype:nexus:0.7.0

The dependency report even suggests that it has low confidence the CPE is correct. It then goes on to refer to the following CVE

http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2014-9389

"Directory traversal vulnerability in Sonatype Nexus OSS and Pro before 2.11.1-01 allows remote attackers to read or write to arbitrary files via unspecified vectors."

This CVE appears to be on Nexus OSS, not jjwt.

The workaround appears to be to add jjwt/0.7.0 to an exclusion list (found in the html report).

Not a show-stopper once I read the find print, but certainly gave me pause for a few minutes.

lhazlewood:
@mkempster is there anything you think we need to do for JJWT to avoid this for others in the future?

mkempster:
A couple small modifications to the MANIFEST.MF seem to prevent jjwt from being matched up to the wrong project cpe.

```
diff MANIFEST.MF.0.7.0 MANIFEST.MF
3,4c3
< Bundle-Description: Sonatype helps open source projects to set up Mave
<  n repositories on https://oss.sonatype.org/
---
> Bundle-Description: JJWT
27,28c26
< Implementation-URL: http://nexus.sonatype.org/oss-repository-hosting.h
<  tml/jjwt
---
> Implementation-URL: https://github.com/jwtk/jjwt

```

I manually made those changes to the MANIFEST.MF, repacked the jar and jjwt disappeared from the list.


For an individual project that needs a workaround, the dependency-check tool suggests adding the jjwt jar to the false-positives-exclusion file with this xml snippet

```
    <suppress>
        <notes><![CDATA[
   file name: jjwt-0.7.0.jar
   ]]></notes>
        <gav regex="true">^io\.jsonwebtoken:jjwt:.*$</gav>
        <cpe>cpe:/a:sonatype:nexus</cpe>
    </suppress>
```

The dependency-check tool's description page https://jeremylong.github.io/DependencyCheck/general/internals.html

says the following

> the JarAnalyzer will collect information from the Manifest, pom.xml, and the package names within the JAR files scanned

mkempster:
Setting just the project.url appears to be sufficient.

[...]

PR #272, opened 2017-10-20
Add project.url to satisfy the dependency-check-maven plugin

mkempster:
The MANIFEST.MF contains an Implementation-URL key, that comes from
project.url if defined. If not defined, jjwt's url comes from sonatype
which seems to falsely associate it with a sonatype/nexus CVE and
fails the project build.

https://github.com/jwtk/jjwt/issues/234


lhazlewood:
Thanks for the PR!

PR #295, opened 2018-01-04
Add Proguard setup section on README

roskoff:
Add Proguard rules for Android projects so it can be referenced right away.


lhazlewood:
Thanks for the PR!

PR #301, opened 2018-02-10
Use java8 java.util.Base64 instead of javax.xml.bind.DatatypeConverter..

dachristensen:
… to avoid ClassDefNotFound runtime errors when invoking via jdk9 or newer without the --add-module javax.xml.bind

This is still compatible on Java8 as the java.util.Base64 classes were added in JDK8. The same classes are also available on Android which eliminates the need for the special AndroidBase64 implementation which uses android.util.Base64. This in turn eliminated the need for the TextCodecFactory. I also deleted the associated tests for TextCodecFactory and Base64Codec (which tested a method that no longer exists).


dachristensen:
Argh, I somehow missed that you were still supporting java7.

Issue #291, opened 2017-12-21
ExpiredJwtException logs wrong date


mbeattie:
The ExpiredJwtException uses a SimpleDateFormatter in the ISO8601 format but doesn't set the timezone to UTC so the string it returns ends with a "Z" indicating it's UTC but the time is parsed in localtime.

The comparisons seem to be correct, just the message is not.

azagniotov:
+1

I experienced a similar issue today:
```
2018-03-12 16:24:47,997 ERROR [pool-13-thread-23] ..... Error details: JWT must not be accepted before 2018-03-12T16:25:19Z. Current time: 2018-03-12T16:24:47Z, a difference of 1003 milliseconds.  Allowed clock skew: 30000 milliseconds.
```
There are `32000` milliseconds difference between the two dates in the log, but the log message shows a difference of `1003`...


Probably has to do with the following code in `DefaultJwtParser`:
```
 Date exp = claims.getExpiration();
            if (exp != null) {

                long maxTime = nowTime - this.allowedClockSkewMillis;
                Date max = allowSkew ? new Date(maxTime) : now;
                if (max.after(exp)) {
                    sdf = new SimpleDateFormat(ISO_8601_FORMAT);
                    String expVal = sdf.format(exp);
                    String nowVal = sdf.format(now);

                    long differenceMillis = maxTime - exp.getTime();

                    String msg = "JWT expired at " + expVal + ". Current time: " + nowVal + ", a difference of " +
                        differenceMillis + " milliseconds.  Allowed clock skew: " +
                        this.allowedClockSkewMillis + " milliseconds.";
                    throw new ExpiredJwtException(header, claims, msg);
                }
            }
```

@lhazlewood in general: is there any plans to move away from using `java.util.Date` in favor of classes from `java.time.*`?

azagniotov:
I see there is another issue raised regarding the same question: https://github.com/jwtk/jjwt/issues/286

lhazlewood:
This has been fixed and released in 0.10.0.

Issue #286, opened 2017-11-30
Support java.time Instant and java.time Clock


Marthym:
It should be a great thing to be able to support java.time.Instant and java.time.Clock instant of Date and io.jsonwebtoken.Clock.

``` java
Jwts.builder()
       .setSubject(user.username())
       .setId(tokenId.toString())
       .setExpiration(Instant.now().plus(expiration))
       .setIssuedAt(Date.from(now))
       .addClaims(claimsBldr.build())
       .signWith(SignatureAlgorithm.HS512, signingKey)
       .compact();
```

```
Jwts.parser()
       .setSigningKey(signingKey)
       .setClock(Clock.systemUTC())
       .parseClaimsJws(token.replace(PREFIX, ""))
       .getBody();
```

lhazlewood:
This depends on #308

lhazlewood:
Closing as duplicate of #235.  Please #235 for work/discussion.

Issue #308, opened 2018-03-07
Abandon java 7 in favour of Java 8 as a baseline


lazystone:
That will allow to fix #282 without additional dependencies.

aharonha:
This Is a great idea, although some legacy dependent projects might break.
This calls for new versioning or forking.

aadrian:
> some legacy dependent projects might break.

@aharonha what projects do you mean?

aharonha:
> @aharonha what projects do you mean?

ones that we can't see ;-)

aharonha:
> Just working with `java.time` package would do the job.

e.g. this PR #311

azagniotov:
@aharonha moving to Java `8` is definitely a breaking & non-backwards compatible change. I imagine the `major` version would need to be incremented, i.e.: `10.0.0`

lhazlewood:
Because of a backwards-incompatible breaking change, semver requires this to wait until at least `1.0.0`.  Assigning that milestone and we'll deal with it at that time.

UkrainianCitizen:
I can help with the migration to the java.time.* or to Java 8 at all. @lhazlewood do you need such a help?

DanFTRX:
According to this: https://developer.android.com/about/dashboards/ more than half of all android devices still ran on Java 7 last year. (Java 8 is on devices with API 24+). 

A lot of people seem to forget that Android Java has a different release schedule than regular Java. While regular Java may be at version 11, Android Java is still barely jumping into Java 8 compatability.

mayonesa:
Supporting Java 7 project here.

ifzm:
+1

UkrainianCitizen:
> 
> 
> Supporting a Java 7 project w/ this library.

Migrate to 8, it is 2019 already, for God's sake. Anyway, you can just use an older version. 
@lhazlewood when this will be implemented? I guess it would be an easier and smoother transition if it isn't done all at once. It would be feasible, for instance, firstly move to `java.time` because I guess it is the easiest part of this transition so to speak, break the transition into several parts as well as this issue. Only after that transition to streams etc. will follow. What do you think about that? I can do that.

DanFTRX:
@UkrainianCitizen Dropping compatability for nearly half a userbase(Probably closer to 1/3 now but google apparently isn't updating it's numbers anymore) is not really a valid option for people.

UkrainianCitizen:
> 
> 
[...]

PR #311, opened 2018-03-13
Added support for java.time Instant instead of java.util.Date

azagniotov:
This PR is trying to address what has been requested in https://github.com/jwtk/jjwt/issues/286 & indirectly in https://github.com/jwtk/jjwt/issues/291

### Summary
- Added support for `java.time.Instant` instead of `java.util.Date`
- Made `ObjectMapper` `static final` in `DefaultJwtParser`.

### Highlights
- Added dependency on Jackson's `jackson-datatype-jsr310`, which brings in support for Java 8 date/time types (specified in `JSR-310` specification)
- All occurrences of `java.util.Date` have been replaced with `java.time.Instant` in prod & test code
- In `DefaultJwtParser`, an instance of `JavaTimeModule` has been registered to bring in deserialization support for parsing timestamp into an `Instant` & the `ObjectMapper` has been made `static final`, since I already touched that code. This also addressed https://github.com/jwtk/jjwt/pull/203 (https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance)
- In `DefaultJwtBuilder`, a serializer for `Instant` has been registered with `ObjectMapper` to serialize `Instant` as a timestamp in seconds as per JWT RFC
- JavaDocs have been updated where appropriate 

### Please note
This PR removes support for building the library against Java 7 as it leverages `java.time.*` classes which were added in Java 8. I do not know what the maintainers of `jjwt` have in mind in terms of sunsetting support for Java 7, so I am not sure if my PR is the right direction

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ dist: trusty
 sudo: required
 language: java
 jdk:
-  - openjdk7
   - oraclejdk8
   - oraclejdk9
 
--- a/pom.xml
+++ b/pom.xml
@@ -109,6 +109,12 @@
             <version>${jackson.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+
         <!-- Optional Dependencies: -->
         <dependency>
             <groupId>org.bouncycastle</groupId>
--- a/src/main/java/io/jsonwebtoken/Claims.java
+++ b/src/main/java/io/jsonwebtoken/Claims.java
@@ -15,7 +15,7 @@
  */
 package io.jsonwebtoken;
 
-import java.util.Date;
+import java.time.Instant;
 import java.util.Map;
 
 /**
@@ -111,13 +111,13 @@ public interface Claims extends Map<String, Object>, ClaimsMutator<Claims> {
      *
      * @return the JWT {@code exp} value or {@code null} if not present.
      */
-    Date getExpiration();
+    Instant getExpiration();
 
     /**
      * {@inheritDoc}
      */
     @Override //only for better/targeted JavaDoc
-    Claims setExpiration(Date exp);
+    Claims setExpiration(Instant exp);
 
     /**
      * Returns the JWT <a href="https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-25#section-4.1.5">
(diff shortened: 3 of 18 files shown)


lmcgrath:
@lhazlewood this is a great addition, I hope to see it in a release soon.

encee22:
@lhazlewood Any thoughts on this PR?

bodawei:
This seems like a good idea to me.  Given the comment on line 48 of DefaultJwtBuilder.java, should it just include a link to the jwt standard?  I see a comment like that, and if I care, I'd like to actually be able to go see what the wording is.

lhazlewood:
JDK 7 support is still necessary - lots of frameworks and tools that use JJWT may not yet be on JDK 8.  I don't know if/when we'll drop JDK 7, but if we do, it'll likely be on the 1.0 final release and not before then since such a change requires a semver major version increment.

RDevet:
@lhazlewood Now that Java 11 is out, I strongly believe you should merge this PR.  Java 7 is now legacy.  Those still using Java 7 can continue to use the previous versions of JJWT.

azagniotov:
I am happy to resolve the conflicts

devEffort:
👍

aniket-patil:
This change is really important to have and I would strongly recommend merging this in.

DanFTRX:
According to this: https://developer.android.com/about/dashboards/  more than half of all android phones still ran on Java 7 last year.

lhazlewood:
As @DanFTRX has said, JDK 7 is still useful.  We'll support it up to the 1.0 final release.

Is there a particular reason why this is important to people when `Date.from(instant)` will work?  I'm just trying to understand if there are other issues beyond that.

RDevet:
@lhazlewood Now that Java 13 is out, I strongly believe you should merge this PR. Java 7 is now a billion years old. Those still using Java 7 can continue to use the previous versions of JJWT.

lhazlewood:
@RDevet as has been discussed in #308, per [semantic versioning](https://semver.org) - which we absolutely will not break since thousands of projects rely on this determinism - this can only be done when we increment the major version number to 1.0.  And for 1.0, we will not release it without the necessary JWE support.  Feel free to follow along with #308 if you like, but this decision isn't being made lightly - believe me, I'd *love* to just use Java 8+ APIs.

Additionally, I would add a friendly caution to not get caught in the trap that Java 13 is oh-so-much-more advanced than JDK 8 or 11.  That probably would have been true when JDK releases were feature-based.  But they're not any longer - they're time-based - every 6 months, and IMO (and many others' opinions), time-based releases are misguided because what *really* matters is semantic versioning for deterministic software upgrades.   Time-based releases just for the sake of meeting a time schedule - as opposed to something stable that people can depend on at a foundational level - is misguided at best IMO.

lhazlewood:
@azagniotov I know this is old, and a lot of changes have been made since you made this PR, but we'd like to include your work in `1.0.x` branch if you're still able.  Would you like to rebase against the `1.0.x` branch?  If so I can merge it in the coming days.  If not, that's totally ok, I can refactor as needed, but I wanted to let you decide.  Please advise.  Thanks!

azagniotov:
@lhazlewood I appreciate you getting in touch about this. I do not think I can get back to rebasing this change anytime soon, so please feel free to refactor as you see fit. Once again, I appreciate you asking me about this.

lhazlewood:
Thanks @azagniotov!

Issue #317, opened 2018-03-28
io.jsonwebtoken.impl.Base64Codec does not work with Java 9 


floragunncom:
<pre>
java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
  at io.jsonwebtoken.impl.Base64Codec.decode(Base64Codec.java:26) ~[jjwt-0.9.0.jar:0.9.0]
</pre>

Suggestion is to use replace javax.xml.DatatypeConverter with java.util.Base64

RyanBard:
See #297 and #282 for details on this (this ticket is a duplicate of at least 1 of those).  For an idea of a fix and/or workaround, you can look at #282 and its open PR: https://github.com/jwtk/jjwt/pull/283/files

floragunncom:
Its not only about DatatypeConverter its also about the general java 9 readiness of this library

RyanBard:
I didn't run into any other problems other than jaxb when testing in java 9 (check out: https://github.com/RyanBard/jjwt-consumer-example).  Did I miss something?

floragunncom:
@RyanBard you're right, seems its only the DatatypeConverter which interferes with Java 9.
But have not checked Java 10 or 11-EA

floragunncom:
Any chance that we get a new release soon? The last commit in the master branch is 7 month old! So for me jjwt currently looks like a dead project.

lhazlewood:
@floragunncom it's definitely not dead - we've just been insanely busy.  I *really* hope we can get one out in the next 3-4 weeks.

floragunncom:
Great to hear. Let me know if we could help to drive it forward.

getjack:
@floragunncom Can confirm the problem for JDK10 (10.0.0.1). Problems appears on runtime.

lhazlewood:
Closing as a duplicate of #333.  Work will be tracked via that issue and we hope to have that released asap in the coming days.

pradipta:
Facing this on 11

Issue #297, opened 2018-01-09
Remove usage of javax.xml.bind.DatatypeConverter (removed in Java 9) 


gregoryalary:
JDK9 marks the package javax.xml.bind as deprecated for removal.

The current use results in a java.lang.NoClassDefFoundError :

`Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
	at io.jsonwebtoken.impl.Base64Codec.decode(Base64Codec.java:26)
	at io.jsonwebtoken.impl.DefaultJwtBuilder.signWith(DefaultJwtBuilder.java:99)`

dachristensen:
I ran into this as well. I just submitted a pull request that resolves this, but breaks the build for java7 as it leverages java.util.Base64 which was added in Java8. As such, I closed the pull request myself. If jjwt needs to support java7, then I'm not sure of a good way to get around this without adding an external dependency on something like commons-codec.

RyanBard:
You can also fix the problem by explicitly depending on jaxb:
* https://github.com/jwtk/jjwt/issues/282
* https://github.com/jwtk/jjwt/pull/283/files

lhazlewood:
Closing as a duplicate of #333.  Work will be tracked via that issue and we hope to have that released asap in the coming days.  Thanks to @RyanBard for the jaxb optional dependency pointer!

lhazlewood:
This has been fixed and released in 0.10.0.

PR #283, opened 2017-11-18
closes #282: Fix Java9 jaxb problem

RyanBard:
This adds an explicit jaxb dependency to fix consuming projects that build with
and target java 9.

Issue #143, opened 2016-07-07
Consider using a different Base64 decoder


omsmith:
As per https://github.com/auth0/node-jsonwebtoken/issues/208, some users of your library are getting stuck on passing in a `String` as a secret. My initial response was "that's not base64, it should be throwing" however given the decoder used, that doesn't appear to be the case. I expect the experience would be much improved if a proper base64 decoder was used, which complained about invalid base64.

As an example, here's the difference between the current `parseBase64Binary` and Java 8's `Base64`:

``` java
private static void printWithBadDecoder( String in ) {
  System.out.println(Arrays.toString(
    javax.xml.bind.DatatypeConverter.parseBase64Binary( in )
  ));
}

private static void printWithGoodDecoder( String in ) {
  System.out.println(Arrays.toString(
    Base64.getDecoder().decode( in )
  ));
}

public static void main( String[] args ) {
  printWithBadDecoder( "my-secret-token-to-change-in-production" );
  printWithBadDecoder( "mysecrettokentochangeinproduction" );
  printWithBadDecoder( "mysecrettokentochangeinproductio" );

  printWithGoodDecoder( "my-secret-token-to-change-in-production" );
  printWithGoodDecoder( "mysecrettokentochangeinproduction" );
  printWithGoodDecoder( "mysecrettokentochangeinproductio" );
}
```

```
[-101, 43, 30, 114, -73, -83, -74, -119, 30, -98, -38, 28, -123, -87, -32, 122, 41, -23, -82, -121, 110, 114, -40, -88]
[-101, 43, 30, 114, -73, -83, -74, -119, 30, -98, -38, 28, -123, -87, -32, 122, 41, -23, -82, -121, 110, 114, -40, -88]
[-101, 43, 30, 114, -73, -83, -74, -119, 30, -98, -38, 28, -123, -87, -32, 122, 41, -23, -82, -121, 110, 114, -40, -88]
Exception in thread "main" java.lang.IllegalArgumentException: Illegal base64 character 2d
Exception in thread "main" java.lang.IllegalArgumentException: Last unit does not have enough valid bits
[-101, 43, 30, 114, -73, -83, -74, -119, 30, -98, -38, 28, -123, -87, -32, 122, 41, -23, -82, -121, 110, 114, -40, -88]
```

As you can see, the current decoder skips over illegal characters, and then afterward doesn't require it be correctly sized (ignoring any remaining characters).

lhazlewood:
[...]

PR #318, opened 2018-03-29
Using Base64 instead of DatatypeConverter

azagniotov:
This PR addresses issues #317 & https://github.com/jwtk/jjwt/issues/297 and https://github.com/jwtk/jjwt/issues/143 (points out a significant difference between the `DatatypeConverter` & `Base64`)

Replacing usages of `javax/xml/bind/DatatypeConverter` with ` java.util.Base64`

### Please note
This PR removes support for building the library against Java 7 as it leverages `java.util.Base64` which was added in Java 8. I do not know what the maintainers of `jjwt` have in mind in terms of sunsetting support for Java 7, so I am not sure if my PR is the right direction. If support for for Java 7 will be dropped, then the issue https://github.com/jwtk/jjwt/issues/308 will be addressed too

As an alternative, the following PR https://github.com/jwtk/jjwt/pull/283 tries to solve this issue by adding a dependency on `jaxb-api`, which  brings in `DatatypeConverter` without dropping the Java 7 build support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ dist: trusty
 sudo: required
 language: java
 jdk:
-  - openjdk7
   - oraclejdk8
   - oraclejdk9
 
--- a/src/main/java/io/jsonwebtoken/impl/Base64Codec.java
+++ b/src/main/java/io/jsonwebtoken/impl/Base64Codec.java
@@ -15,14 +15,20 @@
  */
 package io.jsonwebtoken.impl;
 
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
 public class Base64Codec extends AbstractTextCodec {
 
+    private static final Base64.Decoder BASE64_DECODER = Base64.getDecoder();
+    private static final Base64.Encoder BASE64_ENCODER = Base64.getEncoder();
+
     public String encode(byte[] data) {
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -4,7 +4,6 @@ dist: trusty
 sudo: required
 language: java
 jdk:
-  - openjdk7
   - oraclejdk8
   - oraclejdk9
```
marcospereira: Add `oraclejdk10` and `oraclejdk11` here would be good. We can allow failures for `oraclejdk11` by adding the following configuration:

```yaml
matrix:
  allow_failures:
    - jdk: oraclejdk11
```

lhazlewood:
JDK 7 will be dropped in the 1.0 final release, but until then, I think we need to continue support for JDK 7.

Based on this, are you willing to re-issue the PR and pull in the `jaxb` lib as a dependency for JDK 7 environments?

azagniotov:
@lhazlewood I can do that. Thanks

lhazlewood:
@azagniotov please disregard my previous comment - check out #333 for the latest on how we're trying to solve this.

lhazlewood:
Closing in favor of #341.  Thanks for your continued help and review on this @azagniotov !

PR #341, opened 2018-07-08
Deterministic Base64 Behavior

### LINKED ISSUE #333: Deterministic Base64 support
This issue is being created to supersede similarly reported issues and represent a canonical issue to enable the following:

1. Check to see if `java.util.Base64` is available at runtime using `Classes.isAvailable` (or similar), and if so, use that.  This is the [best/fastest one available](http://java-performance.info/base64-encoding-and-decoding-performance/) to most default environments and should be used whenever possible.
    
    `java.util.Base64` is available on [JDK >= 8](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html) and [Android level >= 26](https://developer.android.com/reference/java/util/Base64)

2. If `java.util.Base64` is not available:
    
    1. If the runtime enviornment is Android, use [android.util.Base64](https://developer.android.com/reference/android/util/Base64).  This is available on Android API level >= 8 (i.e. >= Android 2.2, "Froyo")
    
    2. Otherwise, assume < JDK8, and use `javax.xml.bind.DatatypeConverter`.  JDK 7 has this included automatically.

3. In any case, enable `JwtBuilder` and `JwtParser` methods that allow a user to specify their own Base64 `TextCodec` if/when desired, e.g. `Jwts.parser().setBase64Codec(myBase64Codec)`   This last option would allow the user to enable Base64 support in any environment that doesn't match 1 and 2 above.

4. Ensure any codec implementation will throw exceptions on invalid inputs when possible.  See #143 as an example.  Exceptions should be thrown when possible to indicate invalid input, wrapping/propagating the original codec exception.

Implementation note:  Perhaps instead of using `javax.xml.bind.DatatypeConverter` - which doesn't always fail on invalid inputs (see #143), we should embed the Commons Codec or MigBase64 implementation.

lhazlewood:
Fixes #333 

- Implemented new Base64 encoder forked from MigBase64 to guarantee deterministic behavior on all JDK and Android platforms
- Allowed pluggable Encoder/Decoder for JWT building and parsing via new `Encoder` and `Decoder` and `JwtBuilder#base64UrlEncodeWith` and `JwtParser#base64UrlDecodeWith` methods respectively
- Added tests for all new code to retain 100% code coverage
- Enabled `oraclejdk10` and `openjdk10` builds in TravisCI
- Replaced gmaven plugin with gmavenplus to work on JDK >= 9
- Upgraded surefire and failsafe plugins to 2.22.0 to ensure build works on JDK >= 10
- Ensured JavaDoc linter wouldn't fail the build for JDK >= 8 (was previously only 1.8)

### DIFF
--- a/src/main/java/io/jsonwebtoken/JwtBuilder.java
+++ b/src/main/java/io/jsonwebtoken/JwtBuilder.java
@@ -15,6 +15,8 @@
  */
 package io.jsonwebtoken;
 
+import io.jsonwebtoken.codec.Encoder;
+
 import java.security.Key;
 import java.util.Date;
 import java.util.Map;
@@ -345,11 +347,36 @@ public interface JwtBuilder extends ClaimsMutator<JwtBuilder> {
      * <p>This is a convenience method: the string argument is first BASE64-decoded to a byte array and this resulting
      * byte array is used to invoke {@link #signWith(SignatureAlgorithm, byte[])}.</p>
      *
+     * <h4>Deprecation Notice: Deprecated as of 0.10.0, will be removed in 1.0.0</h4>
+     *
+     * <p>This method has been deprecated because the {@code key} argument for this method can be confusing: keys for
+     * cryptographic operations are always binary (byte arrays), and many people were confused as to how bytes were
+     * obtained from the String argument.</p>
+     *
+     * <p>This method always expected a String argument that was effectively the same as the result of the following
+     * (pseudocode):</p>
+     *
+     * <p>{@code String base64EncodedSecretKey = base64Encode(secretKeyBytes);}</p>
+     *
+     * <p>However, a non-trivial number of JJWT users were confused by the method signature and attempted to
+     * use raw password strings as the key argument - for example {@code signWith(HS256, myPassword)} - which is
+     * almost always incorrect for cryptographic hashes and can produce erroneous or insecure results.</p>
+     *
+     * <p>See this
+     * <a href="https://stackoverflow.com/questions/40252903/static-secret-as-byte-key-or-string/40274325#40274325">
+     * StackOverflow answer</a> explaining why raw (non-base64-encoded) strings are almost always incorrect for
+     * signature operations.</p>
+     *
+     * <p>Finally, please use the {@link #signWith(SignatureAlgorithm, Key)} method, as this method and the
+     * {@code byte[]} variant will be removed before the 1.0.0 release.</p>
+     *
      * @param alg                    the JWS algorithm to use to digitally sign the JWT, thereby producing a JWS.
      * @param base64EncodedSecretKey the BASE64-encoded algorithm-specific signing key to use to digitally sign the
      *                               JWT.
      * @return the builder for method chaining.
+     * @deprecated as of 0.10.0 - use {@link #signWith(SignatureAlgorithm, Key)} instead.
      */
+    @Deprecated
     JwtBuilder signWith(SignatureAlgorithm alg, String base64EncodedSecretKey);
 
     /**
@@ -386,6 +413,18 @@ public interface JwtBuilder extends ClaimsMutator<JwtBuilder> {
      * @since 0.6.0
      */
     JwtBuilder compressWith(CompressionCodec codec);
+
+    /**
+     * Perform Base64Url encoding with the specified Encoder.
+     *
+     * <p>JJWT uses a spec-compliant encoder that works on all supported JDK versions, but you may call this method
+     * to specify a different encoder if you desire.</p>
+     *
+     * @param base64UrlEncoder the encoder to use when Base64Url-encoding
+     * @return the builder for method chaining.
+     * @since 0.10.0
+     */
+    JwtBuilder base64UrlEncodeWith(Encoder<byte[], String> base64UrlEncoder);
 
     /**
      * Actually builds the JWT and serializes it to a compact, URL-safe string according to the
--- a/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
+++ b/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
@@ -17,12 +17,20 @@ package io.jsonwebtoken.impl;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import io.jsonwebtoken.*;
+import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.CompressionCodec;
+import io.jsonwebtoken.Header;
+import io.jsonwebtoken.JwsHeader;
+import io.jsonwebtoken.JwtBuilder;
+import io.jsonwebtoken.JwtParser;
+import io.jsonwebtoken.Jwts;
+import io.jsonwebtoken.SignatureAlgorithm;
+import io.jsonwebtoken.codec.Decoder;
+import io.jsonwebtoken.codec.Encoder;
 import io.jsonwebtoken.impl.crypto.DefaultJwtSigner;
 import io.jsonwebtoken.impl.crypto.JwtSigner;
 import io.jsonwebtoken.lang.Assert;
 import io.jsonwebtoken.lang.Collections;
-import io.jsonwebtoken.lang.Objects;
 import io.jsonwebtoken.lang.Strings;
 
 import javax.crypto.spec.SecretKeySpec;
@@ -39,10 +47,18 @@ public class DefaultJwtBuilder implements JwtBuilder {
     private String payload;
 
     private SignatureAlgorithm algorithm;
-    private Key                key;
-    private byte[]             keyBytes;
+    private Key key;
+
+    private Encoder<byte[], String> base64UrlEncoder = Encoder.BASE64URL;
 
     private CompressionCodec compressionCodec;
+
+    @Override
+    public JwtBuilder base64UrlEncodeWith(Encoder<byte[], String> base64UrlEncoder) {
+        Assert.notNull(base64UrlEncoder, "base64UrlEncoder cannot be null.");
+        this.base64UrlEncoder = base64UrlEncoder;
+        return this;
+    }
 
     @Override
     public JwtBuilder setHeader(Header header) {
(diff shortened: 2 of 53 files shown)


lhazlewood:
@RyanBard and @azagniotov - I'd appreciate a code review if you get a chance!

src/test/groovy/io/jsonwebtoken/codec/impl/Base64DecoderTest.groovy
```diff
@@ -0,0 +1,22 @@
...
+class Base64DecoderTest {
+
+    @Test(expected = IllegalArgumentException)
+    void testDecodeWithNullArgument() {
+        new Base64Decoder().decode(null)
+    }
+
+    @Test
+    void testDecode() {
+        String encoded = 'SGVsbG8g5LiW55WM' // Hello 世界
```
RyanBard: I would add in the the RFC test vectors also (I'm not sure where the best place to put them would be among the various encode/decode tests in this PR): https://tools.ietf.org/html/rfc4648#page-12

```
   BASE64("") = ""

   BASE64("f") = "Zg=="

   BASE64("fo") = "Zm8="

   BASE64("foo") = "Zm9v"

   BASE64("foob") = "Zm9vYg=="

   BASE64("fooba") = "Zm9vYmE="

   BASE64("foobar") = "Zm9vYmFy"
```

It would also be nice to cook up some examples to distinguish the differences between the url-safe one also.

src/main/java/io/jsonwebtoken/JwtBuilder.java
```diff
@@ -345,11 +347,36 @@
      * <p>This is a convenience method: the string argument is first BASE64-decoded to a byte array and this resulting
      * byte array is used to invoke {@link #signWith(SignatureAlgorithm, byte[])}.</p>
      *
+     * <h4>Deprecation Notice: Deprecated as of 0.10.0, will be removed in 1.0.0</h4>
```
RyanBard: I notice the deprecation comments say 0.10.0, but this PR seems to be targeting 0.9.1-SNAPSHOT.

If this isn't a mistake, do you mind giving me a brief explanation of your branching/release process (high level enough that it's no more than 2-3 min of your time)?

src/test/groovy/io/jsonwebtoken/codec/impl/Base64DecoderTest.groovy
lhazlewood: Code coverage reports 100% of Base64's implementation covered, for all statements and branches.

That said, definitely happy to add as many test vectors as we can reliably trust :)

src/main/java/io/jsonwebtoken/JwtBuilder.java
lhazlewood: Ah, the PR targets master and master *should* be at `0.10.0-SNAPSHOT` not at `0.9.1-SNAPSHOT` as it currently is.  I need to change that - thanks for the catch.

Technically the deprecation notices themselves can target any patch release (0.9.1, 0.9.2, etc) because they're just JavaDoc changes and the signatures of all _non-internal/private_ API classes/methods are not changing.  That said, it was too much of a hassle to create a separate PR and release just for that so I'm bundling those notifications in the 0.10.0 release.

The other changes associated with this PR have to be 0.10.0 because [semver](https://semver.org/) requires a minor revision increase when new public APIs are introduced (e.g. `Encoder`, `Decoder`, new public methods on `JwtBuilder` and `JwtParser` etc).

So, based on semver, the way I manage all my git repos is:

`master` always reflects the next minor point release baseline (i.e. 'SNAPSHOT' in maven terminology) - 0.10.0-SNAPSHOT, 0.11.0-SNAPSHOT, 1.0.0-SNAPSHOT, etc) and contains all PRs/fixes/whatever pending the Maj.min.0 release.

If a bugfix or patch release is warranted for any `Maj.min.0` release, a branch is created from the `Maj.min.0` release tag with a `MAJ.min.x` naming scheme (suffix is the actual string literal `.x`)

For example, if 0.9.0 has been released, and we need to get a patch out (e.g. upgrading Jackson as I did earlier this week), I create a `0.9.x` branch from the `0.9.0` release tag. and that branch reflects `0.9.1-SNAPSHOT` and contains merged PRs awaiting release.

When 0.9.1 is about to be released, the pending changes also get merged to master (to ensure they're always in all future Maj.min releases).  Then 0.9.1 is released and then the 0.9.x branch reflects `0.9.2-SNAPSHOT` in its pom.xml.

At this point we can either keep the `0.9.x` branch around if we think there might be further patch releases (`0.9.2`, `0.9.3`, etc), or just delete it.  If we delete it and a `0.9.2` release is warranted, we would re-create the `0.9.x` branch based on the `0.9.1` tag.

I tend to prefer to delete `.x` branches after a patch release to keep things clean unless I know there is an immediate need for another follow-up patch very soon.

So in summary:

1. `master` always reflects the staged changes about to go out in the next `Maj.min.0` release and should always reflect `Maj.min.0-SNAPSHOT` in its pom.xml.

2. `Maj.min.x` branches (suffixed with the string literal '`x`') always reflect staged changes about to go out in the next Maj.min.patch release and should always reflect `Maj.min.patch-SNAPSHOT` in its pom.xml.

3. After cutting a `Maj.min.patch` release, those changes should also be merged into master (except for pom.xml changes that violate # 1).

I hope this helps!  Does that make sense?

src/test/groovy/io/jsonwebtoken/codec/impl/Base64DecoderTest.groovy
RyanBard: I wrote a quick program to generate printable ascii chars (it's a decent start at least, and we already have at least a few unicodes in the existing tests):

```
"special: [\n \t], ascii[32..126]: [ !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~]\n"
```

I manually tinkered with it a little, so if my base64 encoding is off a little, it's probably me.  We should also dump a \r in there (I don't trust myself making a \r in vim and didn't write this out to a file in code like I probably should have).

Linux's base64 of it:

```
c3BlY2lhbDogWwogXHRdLCBhc2NpaVszMi4uMTI2XTogWyAhIiMkJSYnKCkqKywtLi8wMTIzNDU2Nzg5Ojs8PT4/QEFCQ0RFRkdISUpLTE1OT1BRUlNUVVZXWFlaW1xdXl9gYWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXp7fH1+XQo=
```

(again, subtle details might be wrong, so verify independent of my manual tinkering -- manual tinkering being where I added slashes b/c I printed the combination of ascii rather than writing to a file)

I'll write it to a file properly in a bit (going to eat lunch rq).

src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
```diff
@@ -294,25 +301,19 @@ public String compact() {
 
         String base64UrlEncodedHeader = base64UrlEncode(jwsHeader, "Unable to serialize header to json.");
 
-        String base64UrlEncodedBody;
+        byte[] bytes;
+        try {
+            bytes = this.payload != null ? payload.getBytes(Strings.UTF_8) : toJson(claims);
```
azagniotov: Should you also trim the `this.payload` and check if it is not empty before getting the bytes?

src/main/java/io/jsonwebtoken/JwtBuilder.java
lhazlewood: I just converted the previous reply to a wiki entry: https://github.com/jwtk/jjwt/wiki/Release-Code-Management for future reference.

azagniotov:
I think the code changes look fine. I agree with @RyanBard to have more test vectors in place

src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
lhazlewood: Not really?  I mean, leading and trailing space could be a valid payload, and we don't want to sanitize people's payloads.  That said, this code isn't new - it's the same as it was before, just moved up a couple of lines, so I wouldn't want to change it unless there was a strong reason (e.g. in the RFC) to do so.  Or at the very least open a new issue to track it and only do it on a 1.0.0 release.

azagniotov: Alright, just wanted to confirm 👍

src/test/groovy/io/jsonwebtoken/codec/impl/Base64DecoderTest.groovy
lhazlewood: @RyanBard I added the suggested test vectors (and duplicated them for Base64Url too).  PR has been updated.

RyanBard: I'm reasonably confident in this one also:

```
input: "special: [\r\n \t], ascii[32..126]: [ !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~]\n"
encoded: "c3BlY2lhbDogWw0KIAldLCBhc2NpaVszMi4uMTI2XTogWyAhIiMkJSYnKCkqKywtLi8wMTIzNDU2Nzg5Ojs8PT4/QEFCQ0RFRkdISUpLTE1OT1BRUlNUVVZXWFlaW1xdXl9gYWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXp7fH1+XQo="
```

I wrote that string to a file and used Linux's base64 util to encode (and then verified the decoding of that encoded str was my original output).

lhazlewood: @RyanBard in the block directly above, is everything to the right of `input: ` Java code? i.e. I can just copy-and-paste from that into a .java source code file?

RyanBard: They were just string literals to dump into an assert, but since your tests are groovy and double quotes are GStrings, it required a little modification (and some replaces for the base64url).

I've PR'd into your branch here: https://github.com/jwtk/jjwt/pull/342

Scavenge what you need from it.

lhazlewood:
Merged - many thanks to you @RyanBard and @azagniotov for your help and review on this one!

PR #342, opened 2018-07-08
tests #333: Add ascii base64 test vector

RyanBard:
* Adds test vector to base64 & base64url for many ascii chars

Signed-off-by: Ryan Bard <john.ryan.bard@gmail.com>


lhazlewood:
Looks good - thanks for the PR!

Issue #335, opened 2018-07-05
Custom JSON marshaller


lhazlewood:
JJWT's internal Jackson marshaller is ... internal.  It can't be modified or replaced by JJWT users.

This issue represents the work to:

1.  Create a new `JsonConverter` (or similarly named) interface.  Instances of this interface perform the JSON work.

2.  Create multiple maven modules in the JJWT project.  The current code would go under a new module, `jjwt-core`. or `jjwt-api` (haven't decided yet).

3.  Create a new maven module `jjwt-orgjson` as the default implementation of the `JsonConverter` interface using the `org.json` APIs (these are only ~ 50k of size compared to `jackson-core`'s 360k, *and* Android already has `org.json` support, so no need to include that dependency in those environments.

4. Create a new maven module `jjwt-jackson` that uses the Jackson API to implement `JsonConverter` for people that like using Jackson.

5. Create a new maven module `jjwt` that is nothing more than a dependency of (`jjwt-core`|`jjwt-api`) and `jjwt-orgjson` to ensure anyone depending on just `jjwt` gets JJWT with the default json parser.

6. Allow specifying the `JsonConverter` on `JwtParser` and `JwtBuilder` instances to allow users to choose whatever JSON implementation they want.

7. If a user doesn't specify a `JwtConverter` go through class/env discovery logic to find and use the first one available, if possible.

lhazlewood:
Module splitting work has been extracted to #348.  This issue depends on #348 completion first.

lhazlewood:
#348 has been completed and merged to master and this issue is now effectively complete.  #354 however may introduce further changes to these modifications.

PR #344, opened 2018-07-09
Pluggable JSON serialization

lhazlewood:
Initial pluggable JSON (de)serialization support with Jackson and org.json as the first implementations. Jackson dependency is still the default, but this will no longer be necessary when #348 is resolved.

#335 Cannot be resolved until changelog and docs are updated.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,7 @@
         <buildNumber>${user.name}-${maven.build.timestamp}</buildNumber>
 
         <jackson.version>2.9.6</jackson.version>
+        <orgjson.version>20180130</orgjson.version>
 
         <!-- Optional Runtime Dependencies: -->
         <bouncycastle.version>1.56</bouncycastle.version>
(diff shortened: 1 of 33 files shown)


lhazlewood:
cc @RyanBard and @azagniotov

pom.xml
```diff
@@ -103,13 +103,15 @@
...
 
     <dependencies>
+
+        <!-- Optional Dependencies: -->
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
+            <scope>compile</scope>
+            <optional>true</optional>
```
RyanBard: This does potentially break peoples' builds (if they weren't explicitly depending on jackson and were getting it transitively from jjwt); think a mention in the CHANGELOG.md makes sense?

lhazlewood: Yeah, most definitely still need to update the changelog before merging this PR.

The other solution that could very well happen for this PR is to break JJWT into multiple modules per #335's description and keep a `jjwt` artifact that is nothing more than a declaration of dependencies on `jjwt-api`, `jjwt-impl` and `jjwt-jackson` (for example).

This guarantees that those depending on the `jjwt` artifact still experience the same behavior.

That said, we don't have to do this and a changelog entry would be fine because JJWT's actual _public_ API wont have backwards-incompatible changes.  That said, the dependency-chain technique could be a nice-to-have that I'm sure would simplify things for people (and us answering questions on StackOverflow ;) ).

RyanBard: That would be how I would do it (slf4j style rather than spring-boot style).

I mean, if I didn't have a unit test exercising the jjwt code in the project I tested with, it wouldn't have even broken the build.  It just would have been a runtime failure.

RyanBard: You also start to run into really weird errors when the versions of classes available don't match up.

Jackson is rough about that.  If you accidentally mix different versions of the jackson libraries (usually due to complex maven hierarchies), you end up with runtime errors when it's actually used.  In a past project, the dependencies were complex and 2.2 or something was mixed w/ 2.7 or so causing problems.  We had unit tests that should have flushed this problem out, however, the test dependencies were different than the runtime dependencies, so you didn't find out until exercising the code in the app container.

lhazlewood: Yeah, I've experienced similar problems with Jackson - not fun.  There's no question that the SLF4J-style is more deterministic, but people do love the spring-boot style.... Hrm...

Issue #320, opened 2018-04-19
Enforcing a specific required algorithm for JwtParser


bcrist:
I was surprised that there isn't already a method like `JwtParser requireAlgorithm(SignatureAlgorithm algorithm)` in JwtParser.  I realize that JwtParser will throw an exception if an attacker tries to get around the signature verification by specifying algorithm "none", but it still seems like if you have control over the algorithm used to create tokens, the best practice for validating them should be to ensure the token matches the expected algorithm exactly.

As a workaround we're currently using something like this:
```
Jwt jwt = Jwts.parser().setSigningKey(...).parse(token);
if (jwt instanceof Jws) {
    if (ALGORITHM != SignatureAlgorithm.forName(((JwsHeader<?>)jws.getHeader()).getAlgorithm())) {
        throw new MalformedJwtException("Algorithm must be " + ALGORITHM);
    }
    ...
} else {
    throw ...
}
```
but that's rather a mouthful compared to:
```
Jwt jwt = Jwts.parser().requireAlgorithm(ALGORITHM).setSigningKey(...).parse(token);
```

lhazlewood:
Nice issue - thanks for taking the time to add this!  I think it's a nice enhancement.

JeremyColton:
Nice addition, When will this update be released?

lhazlewood:
Actually, I don't see why this is necessary?  When you use `parseClaimsJws` as documented for parsing JWSs, you cannot forge an `alg` header and still have signature validation succeed.  You should never use the unqualified `parse` method if you are expecting a specific type of token, e.g. a JWS.

vraravam:
I have a JWT that was created using `SignatureAlgorithm.HS256` - but when I try to parse it, I get an error `SignatureException: JWT signature does not match locally computed signature. JWT validity cannot be asserted`. Is this error related to this feature request?

lhazlewood:
Closing per the author's closure of #346

PR #346, opened 2018-07-10
Required algorithm 

ahopgood:
as per issue [320](https://github.com/jwtk/jjwt/issues/320)
Any feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,11 @@ You can use the following rules on your Android projects (see [Proguard Exclusio
     
     }
     ```
+* **Require Algorithm.** When parsing you can specify that **only** a particular signing algorithm can be used, giving control of algorithm selection to the entity doing the parsing and not allowing the token to choose the algorithm. This provides protection against the "alg":"none" and RS256 as HS256 [attacks](https://auth0.com/blog/a-look-at-the-latest-draft-for-jwt-bcp/).
+
+    ```java
+      
+    ```
 
 ## Currently Unsupported Features
 
--- a/README.md
+++ b/README.md
@@ -222,10 +222,17 @@ You can use the following rules on your Android projects (see [Proguard Exclusio
     
     }
     ```
-* **Require Algorithm.** When parsing you can specify that **only** a particular signing algorithm can be used, giving control of algorithm selection to the entity doing the parsing and not allowing the token to choose the algorithm. This provides protection against the "alg":"none" and RS256 as HS256 [attacks](https://auth0.com/blog/a-look-at-the-latest-draft-for-jwt-bcp/).
+* **Require Algorithm.** When parsing you can specify that **only** a particular signing algorithm can be used, giving control of algorithm selection to the entity doing the parsing and not allowing the token to choose the algorithm. This provides *some* protection against the "alg":"none" and RS256 as HS256 [attacks](https://auth0.com/blog/a-look-at-the-latest-draft-for-jwt-bcp/).
 
     ```java
-      
+    try {
+        Jwts.parser()
+          .requireAlgorithm(SignatureAlgorithm.HS512)
+          .setSigningKey(base64Encodedkey)
+          .parse(compactJws);
+    } catch (SignatureException se) {
+      // We get here if the algorithm the token specifies fails to match the one we require in the parser.
+    }
     ```
 
 ## Currently Unsupported Features
--- a/src/main/java/io/jsonwebtoken/JwtParser.java
+++ b/src/main/java/io/jsonwebtoken/JwtParser.java
@@ -128,6 +128,16 @@ public interface JwtParser {
     JwtParser require(String claimName, Object value);
 
     /**
+     * Ensures that token signatures are signed with the specified {@link SignatureAlgorithm}, this prevents attacks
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -222,6 +222,18 @@ You can use the following rules on your Android projects (see [Proguard Exclusio
     
     }
     ```
+* **Require Algorithm.** When parsing you can specify that **only** a particular signing algorithm can be used, giving control of algorithm selection to the entity doing the parsing and not allowing the token to choose the algorithm. This provides *some* protection against the "alg":"none" and RS256 as HS256 [attacks](https://auth0.com/blog/a-look-at-the-latest-draft-for-jwt-bcp/).
```
lhazlewood: Please do not reference 3rd party URLs other than the JWT RFC's, our own documentation, or de-facto/vetted StackOverflow answers, as 3rd-party sites can be unreliable or even provide incorrect information (The company URL you reference has had their fair share of [security](https://auth0.com/docs/security/bulletins/cve-2018-6874) [vulnerability](https://auth0.com/docs/security/bulletins/cve-2018-6873) [problems](https://thehackernews.com/2018/04/auth0-authentication-bypass.html) and aren't a reliable source of security information IMO).

Additionally, JJWT has never been susceptible to the referenced attack if you use the API as intended (call `parseClaimsJws`).

src/main/java/io/jsonwebtoken/JwtParser.java
```diff
@@ -127,6 +127,16 @@
      */
     JwtParser require(String claimName, Object value);
 
+    /**
+     * Ensures that token signatures are signed with the specified {@link SignatureAlgorithm}, this prevents attacks
+     * based on the token telling the parser to use an inappropriate SignatureAlgorithm
+     * see {@link https://auth0.com/blog/a-look-at-the-latest-draft-for-jwt-bcp/}.
```
lhazlewood: Please remove link.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -78,6 +78,8 @@
 
     private long allowedClockSkewMillis = 0;
 
+    private SignatureAlgorithm signatureAlgorithm;
```
lhazlewood: Please rename to `requiredSignatureAlgorithm`

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -342,6 +350,14 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
             //re-create the jwt part without the signature.  This is what needs to be signed for verification:
             String jwtWithoutSignature = base64UrlEncodedHeader + SEPARATOR_CHAR + base64UrlEncodedPayload;
 
+            if (signatureAlgorithm != null &&
+                    (signatureAlgorithm.compareTo(algorithm) != 0)) {
```
lhazlewood: Please use a `.equals` or `==` comparison, as these are idiomatic for enums.  `.compareTo` is more esoteric.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -342,6 +350,14 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
...
+                String message = "The expected algorithm " + signatureAlgorithm.getValue() + " does not match " +
```
lhazlewood: Let's have this be:

`The 'alg' JWT header value of '" + algorithm.value() + " does not match the parser required value of '" + requiredSignatureAlgorithm.value() + "'.";`

Also the last sentence is unnecessary because we don't want to speculate as to what might have happened and cause undue alarm.

src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -342,6 +350,14 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
...
+                        "the algorithm specified in the JWT header (" + algorithm.getValue() + "). There is a " +
+                        "chance this JWT is trying to force algorithm selection.";
+                throw new SignatureException(message);
```
lhazlewood: I'm not sure `SignatureException` is valid here - ordinarily that was intended to be used during actual signature calculation and verification operations, not header comparisons.  I think `IncorrectHeaderException` (paralleling the `IncorrectClaimException`) might be better.

In other words, just as we have:
```
ClaimJwtException
|-- InvalidClaimException
    |-- MissingClaimException
    |-- IncorrectClaimException
```
we might also need:
```
HeaderJwtException
|-- InvalidHeaderException
    |-- MissingHeaderException
    |-- IncorrectHeaderException
```

lhazlewood:
Also, please see my last comment in the thread for #320 - is this PR even necessary?

lhazlewood:
@ahopgood - any updates? I don't believe this PR is necessary since `parseClaimsJws` does what is expected, no?

ahopgood:
@lhazlewood sorry about the late reply, I believe you're correct, I'll close this PR.

PR #383, opened 2018-08-13
Keys and SignatureAlgorithm helper method parity

### LINKED ISSUE #381: SecretKey algorithm did not equal one of the three required JCA
Hi there,

We have struggled for over 2 hours to get this library to work (for a such a simple task (JWT) as to generate a simple JWT). Not a good start! 

We end up figuring out the final problem to be this exception:

`SecretKey algorithm did not equal one of the three required JCA algorithm names of HmacSHA256, HmacSHA384, or HmacSHA512.`

The collection `PREFERRED_HMAC_ALGS` has `jcaName`s internally as `HmacSHA256` etc. **using camelcase**. However when we generate a key using the library:

`val key = Keys.secretKeyFor(SignatureAlgorithm.HS256)`

The internal jcaName generated in the key is `HMACSHA256`, **all uppercase**. The method `SignatureAlgorithm()` throws an exception [here] https://github.com/jwtk/jjwt/blob/1839ebf5e8004336b018c9923bd09ee7deebdf39/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java#L560

Are we doing something wrong? It can't be this library has such a stupid string comparison bug and nobody else has notice it?!

We are using the version `0.9.0` from `master`.

lhazlewood:
Enhanced test case to verify symmetric logic between the Keys and SignatureAlgorithm helper methods.  Resolves #381

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
+++ b/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
@@ -554,18 +554,20 @@ public enum SignatureAlgorithm {
         if (key instanceof SecretKey) {
 
             SecretKey secretKey = (SecretKey)key;
-            String secretKeyAlg = secretKey.getAlgorithm();
+            int bitLength = io.jsonwebtoken.lang.Arrays.length(secretKey.getEncoded()) * Byte.SIZE;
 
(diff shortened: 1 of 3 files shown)


lhazlewood:
This was merged and release via commit 56b3a71733d8aa9a8c73ca6bb9c744c56d018628

Issue #369, opened 2018-07-30
JSON POJO unmarshalling


lhazlewood:
Per the discussion in #214:

Improve the exception message:

String msg = "Cannot convert existing claim value of type " + value.getClass() + " to desired type " + requiredType + ". JJWT only converts simple Date, Long, Integer, Short and Byte types automatically. Anything more complex is expected to be already converted to your desired type by the JSON Deserializer implementation. You may specify a custom Deserializer for a JwtParser with the desired conversion configuration via the JwtParser#deserializeJsonWith method. See https://github.com/jwtk/jjwt/README.md#whatever for more information.";

Plus:

* JavaDoc for the `Claims#get` method (dunno why it is missing)
* supporting `new JacksonDeserializer(Maps.of("user", User.class))`
* public documentation on marshaling/unmarshaling and how to do it w/ JJWT (which is the `https://github.com/jwtk/jjwt/README.md#whatever` link in the above exception message).

cc @cbudtz

lhazlewood:
Didn't have time to get this in 0.10.0.  We should be able to get this in the next release.

javapapo:
+1 on this . thanks

PR #393, opened 2018-08-24
Add some convenience methods: `getAsString`/`getAsInt`/`getAsDate`/`getAsBoolean`/`getAsMap`

LiYing2010:
Currently, there is only a generic method `get(...)` which need a `Class<T> requiredType` parameter, it's a little inconvenience to use, since we have to pass `requiredType` like `String.class` or `Integer.class`

So it will be more convenience if we have the following methods:
- `getAsString`
-  `getAsInt`, `getAsLong`, `getAsShort`, `getAsByte`
- `getAsDate`
- `getAsBoolean`
- `getAsMap`

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/Claims.java
+++ b/api/src/main/java/io/jsonwebtoken/Claims.java
@@ -171,4 +171,20 @@ public interface Claims extends Map<String, Object>, ClaimsMutator<Claims> {
     Claims setId(String jti);
 
     <T> T get(String claimName, Class<T> requiredType);
+
+    String getAsString(String claimName);
+
+    Integer getAsInt(String claimName);
+
+    Long getAsLong(String claimName);
+
+    Short getAsShort(String claimName);
+
+    Byte getAsByte(String claimName);
+
+    Date getAsDate(String claimName);
+
+    Boolean getAsBoolean(String claimName);
+
+    Map<String, ?> getAsMap(String claimName);
 }
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
@@ -164,4 +164,44 @@ public class DefaultClaims extends JwtMap implements Claims {
 
         return requiredType.cast(value);
     }
+
+    @Override
+    public String getAsString(String claimName) {
+        return get(claimName, String.class);
+    }
+
+    @Override
+    public Integer getAsInt(String claimName) {
+        return get(claimName, Integer.class);
+    }
+
+    @Override
+    public Long getAsLong(String claimName) {
+        return get(claimName, Long.class);
+    }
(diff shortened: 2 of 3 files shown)


lhazlewood:
Respectfully, I don't see any pragmatic difference between:

`getAsString(propName);`
and
`get(propName, String.class);`

It's only a difference of 5 characters, and the alternative methods expand JJWT's API without any extra functionality.

Addtionally, why not `getAsByte`, `getAsByteArray`, `getAsDouble`, etc, etc?  I hope you can see that this kind of explodes, adds more maintenance, and doesn't really add any value beyond the current functionality.

Finally `get(name, class)` must remain to support custom [POJO types when unmarshalling](#369).  Since that method must exist and already does everything you're proposing, I respectfully decline the pull request.

I very much appreciate the effort however!

In the future, you might want to open an issue for work/discussion to vet the idea before going through the effort of a PR.  I'm always grateful for help, but I want to avoid any of us unnecessary work, so let's discuss first.  I hope that helps!

LiYing2010:
I am sorry, I didn't explain this PR more clearly.

In my opinion, the `getAsXXXType()` or `getXXXType()` pattern is better, because the method name explains it's behavior more clearly.
And also, this pattern is widely used in many libraries/frameworks, e.g:
JDBC ResultSet:
https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html

Android: ContentValues
https://developer.android.com/reference/android/content/ContentValues

Java 7 API: JsonObject:
https://docs.oracle.com/javaee/7/api/javax/json/JsonObject.html

Google GSON:
https://google.github.io/gson/apidocs/com/google/gson/JsonElement.html

Grails: AbstractTypeConvertingMap:
http://docs.grails.org/3.2.3/api/grails/util/AbstractTypeConvertingMap.html

Java Reflection: Field
https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Field.html

OpenCV: Algorithm
https://docs.opencv.org/java/2.4.2/org/opencv/core/Algorithm.html

Apache Commons Configuration:
https://commons.apache.org/proper/commons-configuration/javadocs/v1.10/apidocs/org/apache/commons/configuration/Configuration.html

MongoDB Java Driver:
http://mongodb.github.io/mongo-java-driver/3.8/javadoc/org/bson/BasicBSONObject.html




> It's only a difference of 5 characters

For Java, yes, but for Kotlin, we have to use `String::class.java`, it's a little long

> Finally get(name, class) must remain to support custom POJO types when unmarshalling. 

No, I didn't suggest to remove this `get(name, class)` method, I am only proposing to add some helper methods for most frequently used data types

> In the future, you might want to open an issue for work/discussion to vet the idea before going through the effort of a PR

I am sorry. I will do that next time.

LiYing2010:
Mr  @lhazlewood  :
I am sorry, could you please take a look at my comment above?

Issue #399, opened 2018-09-26
JDK 10+ support


gutmox:
I'm using java 9 modules in my project and when I try to use it, I get this compilation error:

```
> Task :auth:compileJava FAILED
error: the unnamed module reads package io.jsonwebtoken.io from both jjwt.api and jjwt.jackson
error: module reactor.core reads package io.jsonwebtoken.io from both jjwt.api and jjwt.jackson
error: module spring.boot.starter reads package io.jsonwebtoken.io from both jjwt.api and jjwt.jackson
error: module jjwt.jackson reads package io.jsonwebtoken.io from both jjwt.api and jjwt.jackson
error: module com.fasterxml.jackson.databind reads package io.jsonwebtoken.io from both jjwt.api and jjwt.jackson
error: module jjwt.impl reads package io.jsonwebtoken.io from both jjwt.api and jjwt.jackson
...
```

My gradle file looks like this:

```
implementation 'io.jsonwebtoken:jjwt-api:0.10.5'
runtime 'io.jsonwebtoken:jjwt-impl:0.10.5','io.jsonwebtoken:jjwt-orgjson:0.10.5'
```

I couldn't make it work or find any workaround to this issue.

lhazlewood:
Wasn't JDK 9 abandoned/deprecated?  They dropped support for it immediately after 9.0.4.  It's not a version JJWT has ever supported and because not even Oracle does anymore, JJWT probably won't either.   JDK 10 and 11 are more appropriate candidates, no?

lhazlewood:
Got it - thanks.  Yeah, we don't build/test on >= JDK 10 just yet, but I'll use this issue to track that work.

mcgin:
The problem here is what is known as a split package problem.  

The package `io.jsonwebtoken.io` is defined in both the jjwt-jackson and jjwt-api modules which is not allowed in J9+.  It can be resolved by moving the two classes in jjwt-jackon to a new package.

`io.jsonwebtoken.io.JacksonDeserializer` -> `io.jsonwebtoken.jackson.JacksonDeserializer`
`io.jsonwebtoken.io.JacksonSerializer` -> `io.jsonwebtoken.jackson.JacksonSerializer`

mcgin:
The same issue applies to the two classes in `jjwt-orgjson`

jaapcoomans:
[...]

PR #400, opened 2018-09-27
Issue https://github.com/jwtk/jjwt/issues/399 solved: repackaging ext…

gutmox:
Solving issue: https://github.com/jwtk/jjwt/issues/399

### DIFF
--- a/extensions/jackson/src/main/java/io/jsonwebtoken/io/serializer/JacksonDeserializer.java
+++ b/extensions/jackson/src/main/java/io/jsonwebtoken/io/serializer/JacksonDeserializer.java
@@ -1,6 +1,8 @@
-package io.jsonwebtoken.io;
+package io.jsonwebtoken.io.serializer;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.jsonwebtoken.io.DeserializationException;
+import io.jsonwebtoken.io.Deserializer;
 import io.jsonwebtoken.lang.Assert;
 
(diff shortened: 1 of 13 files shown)


gutmox:
Yes, you're right, it breaks backward compatibility but I felt "just repackaging" is not a huge deal and it wouldn't bother too much current projects using this library; 

The only alternative I see is distributing also, as it was before, a single artefact, so developers who want to use java modules can just import that single artefact with the whole the project.

lhazlewood:
Yeah, we can't break backwards compatibility.  There are hundreds if not thousands of projects that use JJWT.  But the fix is easy - just deprecate the existing ones and add the new ones to the desired package.  Would you be willing to change the PR to do that?

Julien-Devin:
Hi @gutmox,  @lhazlewood, 

I allow myself to relaunch this PR, because we use this jjwt on few projects, and we are blocked, the new projects are all modular and we have the error : 
_java.lang.module.ResolutionException: Module jjwt.api contains package io.jsonwebtoken.io, module jjwt.jackson exports package io.jsonwebtoken.io to jjwt.api_

Do you have any solution ? If not, a resolution time ?

lhazlewood:
@gutmox I wanted to say thanks again for the PR!  However due to the scope of #496 (with backwards compatible shaded jars), that issue superseded this one because it's a superset of this PR's functionality, so we'll be closing this one.  Thank you!

PR #414, opened 2018-11-22
Added Gson serialization/deserialization Extension

patton73:
I have provided Gson Extension to serialize/deserialize json.
I have also provided tests as in the other extensions.


lhazlewood:
Thank's for the PR!  We'll see if we can get this in the next minor release!

romeara:
This is something I'd be interested in using - if further work is needed before this is ready to be added, please let me know and I'd be happy to help!

DanFTRX:
I think you need to actually add it to the locator: https://github.com/jwtk/jjwt/blob/master/impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathSerializerLocator.java#L32 

And at that point you would probably want https://github.com/jwtk/jjwt/pull/425  to come in first because GSON is a popular library and the current locator method would then be choosing a different serializer without people knowing it.

patton73:
Is there a particular reason that keeps this pull request blocked for about 3 months? And more if https://github.com/jwtk/jjwt/pull/425 is a precondition why this is not included in 0.11.0 milestone?
Thanks.

lhazlewood:
@patton73 no reason other than it has been a lower priority than shoring up the JWE work.  I hope I'll be able to get to this this week.

lhazlewood:
P.S. *Thank you* for the PR!  It has just been crazy lately, so hopefully we can address this asap.

DanFTRX:
@patton73 This still needs to add the relevant lines to https://github.com/jwtk/jjwt/blob/master/impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathSerializerLocator.java  to load the GSON extension when present.

patton73:
> @patton73 This still needs to add the relevant lines to https://github.com/jwtk/jjwt/blob/master/impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathSerializerLocator.java to load the GSON extension when present.

Yes i know. i Thought they could merge this PR earlier. I think it can be done by mantainers. It is only a matter of add another if with right class and package.

lhazlewood:
One significant reason this PR hasn't been merged yet is that the RuntimeClasspathSerializerLocator hasn't been updated to load GSON when present.  Is there a reason why you didn't want to update that?

extensions/pom.xml
```diff
@@ -36,6 +36,7 @@
     <modules>
         <module>jackson</module>
         <module>orgjson</module>
+	<module>gson</module>
```
lhazlewood: Minor formatting fix request.

patton73:
I did not understand this PR was waiting those changes. I will update i as soon as i can.

patton73:
Mmm. It did not compile. it complaints about "Could not find artifact io.jsonwebtoken:jjwt-root:pom:0.11.0-SNAPSHOT in sonatype-snapshots". What should i do?

lhazlewood:
@patton73 I'm digging in to this today - were you able to rebase from master?

patton73:
I did not have problems in rebasing the master.

lhazlewood:
@patton73 but the build is failing: https://travis-ci.org/jwtk/jjwt/builds/559075398?utm_source=github_status&utm_medium=notification

patton73:
Same strange error about sonatype artifacts. I am afraid I do not know what to do.

DanFTRX:
@patton73 @lhazlewood Probably because your pom references [0.11.0-SNAPSHOT](https://github.com/jwtk/jjwt/pull/414/files#diff-905f6582f56a9b724aa26f8467672a09R24) but the root pom in the repo is [0.10.8-SNAPSHOT](https://github.com/jwtk/jjwt/blob/master/extensions/orgjson/pom.xml#L24)

patton73:
I try to modify root pom in the repo to 0.10.8-SNAPSHOT and see what happens.

patton73:
Ok it worked. It still complaints in the jdk8 compilation but the problem is solved.

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java
```diff
@@ -34,6 +34,8 @@
             return Classes.newInstance("io.jsonwebtoken.io.JacksonDeserializer");
         } else if (isAvailable("io.jsonwebtoken.io.OrgJsonDeserializer")) {
             return Classes.newInstance("io.jsonwebtoken.io.OrgJsonDeserializer");
+        } else if (isAvailable("com.google.gson.GsonBuilder")) {
```
DanFTRX: This should be checking for "io.jsonwebtoken.io.GsonDeserializer" not "com.google.gson.GsonBuilder"

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathSerializerLocator.java
```diff
@@ -34,6 +34,8 @@
             return Classes.newInstance("io.jsonwebtoken.io.JacksonSerializer");
         } else if (isAvailable("io.jsonwebtoken.io.OrgJsonSerializer")) {
             return Classes.newInstance("io.jsonwebtoken.io.OrgJsonSerializer");
+        } else if (isAvailable("com.google.gson.GsonBuilder")) {
```
DanFTRX: This should be checking for "io.jsonwebtoken.io.GsonDeserializer" not "com.google.gson.GsonBuilder"

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java
```diff
@@ -34,6 +34,8 @@
...
+            return Classes.newInstance("io.jsonwebtoken.io.GsonDeSerializer");
```
DanFTRX: This should be "io.jsonwebtoken.io.GsonDeserializer" not "io.jsonwebtoken.io.GsonDeSerializer"

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathSerializerLocator.java
patton73: Ok patching.

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java
patton73: Ok patching.

DanFTRX:
@patton73 I think this should do it:  https://github.com/DanFTRX/jjwt/commit/4c89f0263797eec03a8751119b1b4879b2de9c76

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java
patton73: ok fixed.

pom.xml
```diff
@@ -69,7 +69,7 @@
         <pluginRepository>
             <snapshots>
                 <enabled>false</enabled>
-            </snapshots>
+           </snapshots>
```
DanFTRX: Formatting changed?

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java
DanFTRX: Missed DeSerializer -> Deserializer on both this and the line above now.

pom.xml
patton73: Mmm it does not find io.jsonwebtoken.io.GsonSerializer in the groovy test.

patton73: Anyway the typos problem are resolved now.

DanFTRX:
@patton73 I missed a pom, you need to add the gson extension to this:  https://github.com/jwtk/jjwt/blob/master/impl/pom.xml  as a test dependency

impl/pom.xml
```diff
@@ -54,6 +54,11 @@
             <artifactId>jjwt-orgjson</artifactId>
             <scope>test</scope>
         </dependency>
+	<dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-gson</artifactId>
+	   <version>${project.version}</version>
```
DanFTRX: Not quite.  Drop the version tag, and add a test scope like the one above it, and make sure the formatting is correct.

patton73: Excuse me i ude gradle basically. Whats wrong with farmatting? do i need to use tabs?

DanFTRX: You have less spaces than the line above it. so it is misaligned.

pom.xml
DanFTRX: Oof, it looks like you forgot to revert the change to this line.  You accidentally removed one of the spaces..

DanFTRX:
@lhazlewood Aside from the one formatting change, I think its presentable now.  I'm not sure whats up with the JDK9 travis failure, but it doesn't appear to be caused by the changes.

pom.xml
patton73: Hopefully fixed.

DanFTRX:
@patton73 Looks good to me, now its up to @lhazlewood

extensions/pom.xml
DanFTRX: this is a tab, needs be be 8 spaces instead

impl/pom.xml
```diff
@@ -54,6 +54,11 @@
             <artifactId>jjwt-orgjson</artifactId>
             <scope>test</scope>
         </dependency>
+	<dependency>
```
DanFTRX: This is also a tab, change to 8 spaces

impl/pom.xml
```diff
@@ -54,6 +54,11 @@
...
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-gson</artifactId>
+	    <scope>test</scope>
```
DanFTRX: This has a tab and spaces,  make sure it is 12 spaces, no tabs

impl/pom.xml
```diff
@@ -54,6 +54,11 @@
...
+	</dependency>
```
DanFTRX: This is a tab, change to 8 spaces

pom.xml
```diff
@@ -131,6 +132,11 @@
                 <artifactId>jjwt-orgjson</artifactId>
                 <version>${project.version}</version>
             </dependency>
+	    <dependency>
```
DanFTRX: this has a tab and spaces, make sure it is 12 spaces, no tabs

pom.xml
```diff
@@ -90,6 +90,7 @@
 
         <jackson.version>2.9.9.1</jackson.version>
         <orgjson.version>20180130</orgjson.version>
+	<gson.version>2.8.5</gson.version>
```
DanFTRX: This is a tab, change to 8 spaces

pom.xml
```diff
@@ -141,6 +147,11 @@
                 <artifactId>json</artifactId>
                 <version>${orgjson.version}</version>
             </dependency>
+	    <dependency>
```
DanFTRX: this has a tab and spaces, make sure it is 12 spaces, no tabs

extensions/gson/src/main/java/io/jsonwebtoken/io/GsonDeserializer.java
```diff
@@ -0,0 +1,42 @@
+package io.jsonwebtoken.io;
```
lhazlewood: Because of the Java >= 9 module system, separate implementation .jars cannot export the same packages as the base api .jar.  As a result we have to fix this for the Jackson and org.json implementations and that's being tracked under #399 .

But because this gson extension is new code, we can avoid this mistake and do it properly from the start.  As a result, please move these to a `io.jsonwebtoken.gson.io` package.

This is to use a convention moving forward:

`io.jsonwebtoken` prefix + `.<extension-name>.` + `<mirrored suffix in the api module>`

For example, if there is an interface in the API module in the `io.jsonwebtoken.foo` package, and an extension named `bar` wanted to implement it, then:

* the bar implementation would be in package `io.jsonwebtoken.bar.foo`, and
* the .java file would be in `extensions/bar/src/main/java/io/jsonwebtoken/bar/foo`

This works for the Java >= 9 module system because `io.jsonwebtoken.*` is exported from the api .jar, and only `io.jsonwebtoken.gson.*` is exported from the gson .jar.  There is no set intersection of packages between the two (which the JDK Module System requires).

extensions/gson/src/main/java/io/jsonwebtoken/io/GsonSerializer.java
```diff
@@ -0,0 +1,43 @@
+package io.jsonwebtoken.io;
```
lhazlewood: Per the same explanation of `GsonDeserializer`, please move this to `io.jsonwebtoken.gson.io`

extensions/gson/src/main/java/io/jsonwebtoken/io/GsonSerializer.java
```diff
@@ -0,0 +1,43 @@
...
+    public GsonSerializer(Gson gson) {
+        Assert.notNull(gson, "gson cannot be null.");
+        this.gson = gson;
+    }
+
+    @Override
+    public byte[] serialize(T t) throws SerializationException {
+        Assert.notNull(t, "Object to serialize cannot be null.");
+        //Gson never throws any serialization exception 
+        return writeValueAsBytes(t);
```
lhazlewood: Per the other implementations (Jackson and org.json), this call is wrapped in a try/catch block, and if any `Exception` is thrown, it is wrapped and re-thrown as a `io.jsonwebtoken.io.SerializationException`.  To ensure runtime consistency, we need to do that here too, i.e.

```java
@Override
public byte[] serialize(T t) throws SerializationException {
    Assert.notNull(t, "Object to serialize cannot be null.");
    try {
        return writeValueAsBytes(t);
    } catch (Exception e) {
        String msg = "Unable to serialize object: " + e.getMessage();
        throw new SerializationException(msg, e);
    }
}
```

extensions/gson/src/test/groovy/io/jsonwebtoken/io/GsonDeserializerTest.groovy
```diff
@@ -0,0 +1,64 @@
+package io.jsonwebtoken.io
```
lhazlewood: `io.jsonwebtoken.gson.io`

extensions/gson/src/test/groovy/io/jsonwebtoken/io/GsonSerializerTest.groovy
```diff
@@ -0,0 +1,99 @@
+package io.jsonwebtoken.io
```
lhazlewood: `io.jsonwebtoken.gson.io`

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java
```diff
@@ -34,6 +34,8 @@
             return Classes.newInstance("io.jsonwebtoken.io.JacksonDeserializer");
         } else if (isAvailable("io.jsonwebtoken.io.OrgJsonDeserializer")) {
             return Classes.newInstance("io.jsonwebtoken.io.OrgJsonDeserializer");
+        } else if (isAvailable("io.jsonwebtoken.io.GsonDeserializer")) {
```
lhazlewood: `io.jsonwebtoken.gson.io.GsonDeserializer`

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathSerializerLocator.java
```diff
@@ -34,6 +34,8 @@
             return Classes.newInstance("io.jsonwebtoken.io.JacksonSerializer");
         } else if (isAvailable("io.jsonwebtoken.io.OrgJsonSerializer")) {
             return Classes.newInstance("io.jsonwebtoken.io.OrgJsonSerializer");
+        } else if (isAvailable("io.jsonwebtoken.io.GsonSerializer")) {
```
lhazlewood: `io.jsonwebtoken.gson.io.GsonSerializer`

lhazlewood:
@patton73 will you be able to get to this today?  Thoughts?

patton73:
Pheraps this night. Or at max tomorrow morning. I am travelling and no laptop with me.

lhazlewood:
@patton73 understood - enjoy your trip! Thanks!

DanFTRX:
Whitespace checks out.

extensions/gson/src/test/groovy/io/jsonwebtoken/gson/io/GsonSerializerTest.groovy
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Test
+    void testSerialize() {
+        byte[] expected = '{"hello":"世界"}'.getBytes(Strings.UTF_8)
+        byte[] result = new GsonSerializer().serialize([hello: '世界'])
+        assertTrue Arrays.equals(expected, result)
+    }
+
+//THIS IS COMMENTED OUT BECAUSE GSN NEVER THROWS ANY SERIALIZATION EXCEPTION
```
DanFTRX: You have to re-enable this test @patton73 Code coverage requires it.

patton73: Well gson does not lauch any exception do I write a fake test about that?

DanFTRX: The test actually handles that.  It throws an exception by overriding the serialization on line 82

DanFTRX: Just choose a different exception.  ANy would do.  the OrgJson extension uses IllegalArgumentException:  https://github.com/jwtk/jjwt/blob/73f7f9915baccaa3f3143e549647f818606d49ce/extensions/orgjson/src/test/groovy/io/jsonwebtoken/io/OrgJsonSerializerTest.groovy#L35

patton73: If i re-enable this the test will never pass since gson will never throw a serialization exception.

DanFTRX: The following code creates a Mock GSONSerializer that DOES throw an exception.  Now I agree its a bit silly because GSON won't even throw an exception outside of this unit test, but @lhazlewood wanted the try catch there for consistency and to maintain 100% code coverage you need to hit that try catch
```
    @Test
    void testSerializeFailsWithWrappedException() {

        def ex = createMock(Exception)

        expect(ex.getMessage()).andReturn('foo')

        def serializer = new GsonSerializer() {
            @Override
            protected byte[] writeValueAsBytes(Object o) {
                throw ex
            }
        }

        replay ex

        try {
            serializer.serialize([hello: 'world'])
            fail()
        } catch (SerializationException se) {
            assertEquals 'Unable to serialize object: foo', se.getMessage()
            assertSame ex, se.getCause()
        }

        verify ex
    }```

patton73: Just for curiosity, is there a reason behind the use of groovy to test java instead of junit?

DanFTRX:
@lhazlewood No problem, happy to help.

extensions/gson/src/test/groovy/io/jsonwebtoken/gson/io/GsonSerializerTest.groovy
lhazlewood: @patton73 we are using JUnit as the testing library.  We just use Groovy to write the tests - it's faster to write, allows us to ignore Java visibility modifiers to assert values and state, reflection is far easier, mocking things can be simpler, and it's fun to use.

In other words, it's far simpler and more productive to write tests in Groovy, but we still use JUnit to execute the tests and assertions.  HTH!

PR #421, opened 2018-12-28
Extension Packaging

acoburn:
Addresses #399 

I have encountered the same issue reported by @gutmox when trying to use jjwt with Java 11 modules. (The module system doesn't work with split packages)

This PR takes @gutmox's commit in #400 and adds some backwards compatibility and deprecation notices to the original class names.

I would note, however, that this issue won't be resolved until the `io.jsonwebtokens.io` package is no longer exported by the extension modules. This PR would allow you to release a 0.10.6 with the deprecation notice and a path towards desired future use. Then, with release 0.11, you could eliminate the split package issue entirely by removing the classes that are (re-)introduced here.

### DIFF
--- a/extensions/jackson/src/main/java/io/jsonwebtoken/io/JacksonDeserializer.java
+++ b/extensions/jackson/src/main/java/io/jsonwebtoken/io/JacksonDeserializer.java
@@ -1,0 +1,9 @@
+package io.jsonwebtoken.io;
+
+/**
+ * @deprecated Please use io.jsonwebtoken.io.serializer.JacksonDeserializer instead.
+ */
+@Deprecated
+public class JacksonDeserializer<T> extends io.jsonwebtoken.io.serializer.JacksonDeserializer<T> {
+
+}
--- a/extensions/jackson/src/main/java/io/jsonwebtoken/io/JacksonDeserializer.java
+++ b/extensions/jackson/src/main/java/io/jsonwebtoken/io/JacksonDeserializer.java
@@ -1,9 +1,18 @@
 package io.jsonwebtoken.io;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 /**
  * @deprecated Please use io.jsonwebtoken.io.serializer.JacksonDeserializer instead.
  */
 @Deprecated
 public class JacksonDeserializer<T> extends io.jsonwebtoken.io.serializer.JacksonDeserializer<T> {
 
+    public JacksonDeserializer() {
+        super();
+    }
+
+    public JacksonDeserializer(ObjectMapper objectMapper) {
+        super(objectMapper);
+    }
 }
--- a/extensions/jackson/src/main/java/io/jsonwebtoken/io/JacksonSerializer.java
+++ b/extensions/jackson/src/main/java/io/jsonwebtoken/io/JacksonSerializer.java
@@ -1,0 +1,9 @@
+package io.jsonwebtoken.io;
+
+/**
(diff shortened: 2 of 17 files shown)


lhazlewood:
@acoburn the deprecation approach is the ideal one so that existing code doesn't break.  That said, will just re-creating those files in a separate module solve this issue for you?  Will the existing classes not cause problems?

acoburn:
@lhazlewood the issue is that there are three jar files each exporting the `io.jsonwebtoken.io` package. Moving the `jjwt-orgjson` and `jjwt-jackson` serializer code into separate package namespaces will fix this, since there would no longer be any split packages.

lhazlewood:
@acoburn is that the case even though the existing classes still remain?  Will they not cause problems?

I'm assuming not because JJWT won't be instantiating them - i.e. while they will exist in the core jar, they won't be used by JJWT internals, and therefore won't cause problems.  Is this correct?

acoburn:
Leaving the existing classes in place will not work for JDK 11 modules. In order to use JJWT within the JPMS system, those classes from the `io.jsonwebtoken.io` package (i.e. in the `jjwt-jackson` and `jjwt-orgjson` jars) will need to be removed. The implementation in this PR simply suggests a path forward, in that the existing (split-package) classes are first deprecated so that they can be entirely removed at some future point.

lhazlewood:
Got it thanks.

Another question: wouldn't this only be solved in best-practice manner if we had these extensions in their own extension-specific package?

That is:

```
io.jsonwebtoken.io.jackson.JacksonSerializer
io.jsonwebtoken.io.jackson.JacksonDeserializer
```
and
```
io.jsonwebtoken.io.orgjson.JacksonSerializer
io.jsonwebtoken.io.orgjson.JacksonDeserializer
```

as opposed to putting all implementations in the `io.jsonwebtoken.io.serializers` package?

Also - out of curiosity - it seems that the JDK 11 module system pretty much destroys the ability for modules in the same project to use package-protected methods, no?

For example, if I have a package `com.foo.bar` and projectA-jar1 that has an abstract class with package-protected methods, and in another module in the same project, I have projectA-jar2 that creates a subclass of that class and wants to call those package-protected methods, they *have* to live in the same package `com.foo.bar` package per the the Java language spec.

That seems to be a big issue...

lhazlewood:
Also - is there nothing in the JPMS that allows us to write metadata e.g. in a jar's META-INF directory that says "Hey, I know these classes are in here, but ignore them please".  I need to research...

acoburn:
Yes, a best-practice structure would put the code for each of the extension modules into its own extension-specific package, such as you suggest (`io.jsonwebtoken.io.jackson.JacksonSerializer` and `io.jsonwebtoken.io.orgjson.OrgJsonSerializer`).

The package-protected visibility still exists in JDK 11, but only within a given package. And JPMS requires that package boundaries do not cross JAR boundaries (i.e. you can't have multiple JARs accessing the same package-protected methods/classes). Hence modularity. Given JAR-A, if you want any methods/classes to be available to JAR-B, those methods/classes will have to be public.

If you are thinking about how the serializer/deserializer [is loaded](https://github.com/jwtk/jjwt/blob/master/impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java#L33-L39) you may want to consider using the [Java ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) which would provide a greater level of flexibility for how those extensions are loaded.

acoburn:
The issue that this PR addresses has already been resolved

PR #457, opened 2019-06-24

jaapcoomans:
Currently there are several dependencies on class names between modules that go in the reverse direction of the actual dependency. This is because hardcoded class names and reflection are being used. By using `ServiceLoader` instead of the current setup, the dependency points in the right direction at all source levels. This makes `jjwt-api` fully independent and agnostic of `jjwt-impl` and `jjwt-impl` fully independent and agnostic of `jjwt-jackson` and `jjwt-orgjson`. 

It is quite a big change, but I believe it's worth it. It makes it easier to develop API's, implementations and extensions separately. Currently many reported issues are blocked by the fact that jjwt is used on both Android (JDK 1.7) and regular Java (JDK 8+). This PR still is fully JDK 1.7 compliant, yet at the same is a first step to supporting JPMS.


jaapcoomans:
I see I was a little careless and let commit 73cfdc8d slip through on this branch. I'll remove it and create a new pull-request.

PR #458, opened 2019-06-26

jaapcoomans:
Currently there are several dependencies on class names between modules that go in the reverse direction of the actual dependency. This is because hardcoded class names and reflection are being used. By using `ServiceLoader` instead of the current setup, the dependency points in the right direction at all source levels. This makes `jjwt-api` fully independent and agnostic of `jjwt-impl` and `jjwt-impl` fully independent and agnostic of `jjwt-jackson` and `jjwt-orgjson`.

It is quite a big change, but I believe it's worth it. It makes it easier to develop API's, implementations and extensions separately. Currently many reported issues are blocked by the fact that jjwt is used on both Android (JDK 1.7) and regular Java (JDK 8+). This PR still is fully JDK 1.7 compliant, yet at the same is a first step to supporting JPMS.


jaapcoomans:
Strange... the builds for JDK7 and JDK10 failed with the following error:

`Unresolveable build extension: Plugin org.apache.felix:maven-bundle-plugin:3.3.0 or one of its dependencies could not be resolved: The following artifacts could not be resolved: org.apache.felix:maven-bundle-plugin:jar:3.3.0, biz.aQute.bnd:biz.aQute.bndlib:jar:3.3.0, org.easymock:easymock:jar:3.4, org.objenesis:objenesis:jar:2.2, org.apache.felix:org.apache.felix.utils:jar:1.6.0, org.apache.maven:maven-plugin-parameter-documenter:jar:2.2.0: Could not transfer artifact org.apache.felix:maven-bundle-plugin:jar:3.3.0 from/to central (https://repo.maven.apache.org/maven2): Access denied to: https://repo.maven.apache.org/maven2/org/apache/felix/maven-bundle-plugin/3.3.0/maven-bundle-plugin-3.3.0.jar , ReasonPhrase:Forbidden.`

Neither the JDK8 or JDK9 builds failed for this reason, while they're using the exact same plugin. Also, when I manually look up all artifacts mentioned in above error message I can all download them from Maven central. Must have been a glitch or something.

Will close and re-open to trigger a fresh build.

jaapcoomans:
So this time the JDK7 and 10 builds succeeded, but now the JDK9 build failed for the same reason. That's eerily flaky. I unfortunately can't connect the  error nor the flakiness with the changes in the PR.

jaapcoomans:
Turns out more people are experiencing similar issues with Travis and Maven Central: https://travis-ci.community/t/continuous-maven-repo-403/3908. Apparently Travis is currently investigating.

jaapcoomans:
An issue was also reported at Sonatype: https://issues.sonatype.org/browse/MVNCENTRAL-4985. According to the comments on that issue several more IP's are now whitelisted. Will trigger a build again to see if this resolved the issue.

jaapcoomans:
According to the Travis incident report the issue is now resolved. https://www.traviscistatus.com/incidents/mm020jn4nzcn

lhazlewood:
@jaapcoomans I just wanted to let you know I haven't forgotten about this - but it is substantial and I need to dig in to it a little further.  For example, for semantic versioning reasons, we cannot move existing classes or interfaces into other packages - they must remain in the packages where they currently are, and it's not yet clear to me if your PR re-arranges packages yet.

Also, things in the `api` module are strictly to be used by end-users of JJWT and should not expose anything they wouldn't use.  The `Services` class looks like it might be such a thing that shouldn't be exposed (but again, I haven't dug in to it enough).  I'll have more feedback as soon as I can look into it.

Thanks!

lhazlewood:
P.S. I do want to say that this PR does look promising!  I like the idea of the `ServiceLoader` concept!

lhazlewood:
Also, @jaapcoomans have you tested this on Android?  Does `ServiceLoader` and .jar metadata work without any issues on Android?

lhazlewood:
Looks like it has been available in Android since Service Level 9, aka "Gingerbread":

https://developer.android.com/reference/java/util/ServiceLoader

jaapcoomans:
Hi @lhazlewood, thanks for the feedback :) I must admit I haven't tested on Android, as I have no idea how to do that. Is support starting from Gingerbread sufficient in your opinion?

I took the backwards compatibility requirement into account. If I'm not mistaken the PR does not break clients of the API.

lhazlewood:
I think that because Gingerbread onwards has support for `ServiceLoader`, we should be ok.  I'll hopefully be able to add PR review comments soon.  Thanks!

bdemers:
Resolved conflicts and continued work in #496 

Thanks again @jaapcoomans !!

PR #472, opened 2019-07-15
#84 JwtBuilder.setClaims() now accepts `Map<String, ?> claims`

### LINKED ISSUE #84: API signature enhancement for JwtBuilder.setClaims
Currently:

```
JwtBuilder setClaims(Map<String, Object> claims);

```

Would be nice if this looked like:

```
JwtBuilder setClaims(Map<String, ? extends Object> claims);

```

This would allow one to pass a Maps of type `Map<String,String>`, etc into this method without needing to cast.

longtimeago:
fixes #84

This would allow to pass `Map<String,String>`


lhazlewood:
Thank you for the PR!

PR #489, opened 2019-09-17
Idea: Use a ServiceLoader to find the Serializer/Deserializer on the classpath

bdemers:
This allows for the discovery of the Serializer implementation without needing to hardcode the class names in RuntimeClasspathSerializerLocator

**NOTE:** Java 1.8+ we can drop the manual managing of the META-INF/services dir and use a compile-time annotation processor: https://github.com/google/auto/tree/master/service

TODO:
- [ ] fix coverage
- [ ] remove RuntimeClasspathSerializerLocator


bdemers:
@lhazlewood pointed out we already had a similar PR from @jaapcoomans #458
I'll take a look at that one, it looks more complete

PR #490, opened 2019-09-17
Add maven-wrapper and configure CI to use it

bdemers:
NOTE: also disable verbose artifact download messages (~40% of the log lines)


bdemers:
IMHO, the biggest reason is for CI consistency. The version of Maven installed is no longer dependent on what Travis provides (and because of this is less suspectable to any unexpected change of behavior between environments), i.e. we are now pinning the builds to use Maven 3.6.1

By updating to Maven 3.6.1 (we can remove the transfer log spam with the new `-no-transfer-progress` flag), granted there are other ways to install a given version of Maven, this is just the easiest.

The Maven Wrapper also lowers the barrier of entry for some projects.  Assuming you have Java installed you could build any project that uses the wrapper.

PR #495, opened 2019-09-19
Add support for custom type deserialization with Jackson

bdemers:
Adds new constructor to `JacksonDeserializer(Map<String, Class>
claimTypeMap)`, which will enable later calls `Claims.get("key",
CustomType.class)` to work as expectd

Fixes: #369

TODO: 
- [x] When we are happy with this, I'll start a 0.11.0-README branch and add a note about how to use this feature. #500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,12 @@
 
 ### 0.11.0
 
+This minor release:
+
 * Updates the Jackson dependency version to [2.9.10](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9#patches)
 to address three security vulnerabilities in Jackson.
+* Adds support for custom types when deserializing with Jackson. To use configure your parser with `Jwts.parserBuilder().deserializeJsonWith(new JacksonDeserializer(Maps.of("claimName", YourType.class).build())).build()`.
+* Adds `io.jsonwebtoken.lang.Maps` utility class to make creation of maps fluent.
 * Moves JSON Serializer/Deserializer implementations to a different package name.
   - `io.jsonwebtoken.io.JacksonSerializer` -> `io.jsonwebtoken.jackson.io.JacksonSerializer`
   - `io.jsonwebtoken.io.JacksonDeserializer` -> `io.jsonwebtoken.jackson.io.JacksonDeserializer`
--- a/api/src/main/java/io/jsonwebtoken/lang/Maps.java
+++ b/api/src/main/java/io/jsonwebtoken/lang/Maps.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2019 jsonwebtoken.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.jsonwebtoken.lang;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Utility class to help with the manipulation of working with Maps.
+ * @since 0.11.0
+ */
+public final class Maps {
+
+    private Maps() {} //prevent instantiation
+
+    /**
+     * Creates a new map builder with a single entry.
+     * <p> Typical usage: <pre>{@code
+     * Map<K,V> result = Maps.of("key1", value1)
+     *     .and("key2", value2)
+     *     // ...
+     *     .build();
+     * }</pre>
+     * @param key the key of an map entry to be added
+     * @param value the value of map entry to be added
+     * @param <K> the maps key type
+     * @param <V> the maps value type
+     * Creates a new map builder with a single entry.
+     */
+    public static <K, V> MapBuilder<K, V> of(K key, V value) {
+        return new HashMapBuilder<K, V>().and(key, value);
+    }
+
+    /**
+     * Utility Builder class for fluently building maps:
+     * <p> Typical usage: <pre>{@code
+     * Map<K,V> result = Maps.of("key1", value1)
+     *     .and("key2", value2)
+     *     // ...
+     *     .build();
+     * }</pre>
+     * @param <K> the maps key type
+     * @param <V> the maps value type
+     */
+    public interface MapBuilder<K, V> {
+        /**
+         * Add a new entry to this map builder
+         * @param key the key of an map entry to be added
+         * @param value the value of map entry to be added
+         * @return the current MapBuilder to allow for method chaining.
+         */
+        MapBuilder and(K key, V value);
+
+        /**
+         * Returns a the resulting Map object from this MapBuilder.
+         * @return Returns a the resulting Map object from this MapBuilder.
+         */
+        Map<K, V> build();
+    }
+
+    private static class HashMapBuilder<K, V> implements MapBuilder<K, V> {
+
+        private final Map<K, V> data = new HashMap<>();
+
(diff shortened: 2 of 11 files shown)


bdemers:
**NOTE:** this sits on top of `fix-split-package` to avoid merge conflicts (as JacksonDeserializer was moved in that branch)

extensions/jackson/src/test/groovy/io/jsonwebtoken/jackson/io/JacksonDeserializerTest.groovy
```diff
@@ -136,6 +137,6 @@ class JacksonDeserializerTest {
     }
 
     private String base64(String input) {
-        return Base64.encoder.encodeToString(input.getBytes('UTF-8'))
+        return new Base64Encoder().encode(input.getBytes('UTF-8'))
```
lhazlewood: Is there a reason you didn't use `Encoders.BASE64` to avoid instantiation?

lhazlewood: Ah, I see - it's just a test.  Still...

bdemers: I’ll fix that, originally I was using Java.util.base64, but that failed on Java7 :(

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This minor release:
 
   A backward compatibility module has been created `io.jsonwebtoken:jjwt-deprecated`, if you are compiling against 
   these classes directly, otherwise you will be unaffected.
+* Adds support for custom types when deserializing with Jackson. To use configure your parser with `Jwts.parserBuilder().deserializeJsonWith(new JacksonDeserializer(Maps.of("claimName", YourType.class))).build()`.
```
lhazlewood: Are we going to create the `Maps.of` utility method? (would reside, presumably, in `io.jsonwebtoken.lang.Maps`).

bdemers: Yeah, easy enough to add. I was thinking of recommending just using the Java 11 `Map.of`, but I know many of us will be on 8 for a while yet.

api/src/main/java/io/jsonwebtoken/lang/Maps.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Utility class to help with the manipulation of working with Maps.
+ * @since 0.11.0
+ */
+public final class Maps {
```
lhazlewood: Shouldn't this be a builder of sorts? Or in addition?

For example, I consider this a lot nicer than a bunch of parameters in a single method:

`Maps.put(key1, value1).put(key2, value2).put(key3, value3).build()`

(for example).

Thoughts?

bdemers: I'm not against it that 🤔 I just went the Java 11-ish route of `Map.of()`, 
I've done something things with a `.put()` in the past, my concern is Map returns the value type on a `put`, so that might be confusing?

I will say, I like the chained method look better than the overloaded method, mostly because it seems DRYer?

I'm not sure if `Map.of()` will ever see common use (at least for a while)

lhazlewood: ah, right - instead of `put`, then maybe `of` or `with` or `and`.

lhazlewood: I like this:

`Maps.of("foo", "whatever").and(key2, value2).and(key3, value3)...`

Maps.of creates a builder instance, puts the first key/value and returns the builder.  And just adds the key/value and returns the builder.  Looks nice and fluent.

Thoughts?

bdemers: I like that!

impl/src/main/java/io/jsonwebtoken/impl/DefaultClaims.java
```diff
@@ -22,6 +22,13 @@
 
 public class DefaultClaims extends JwtMap implements Claims {
 
+    private static final String CONVERSION_ERROR_MSG = "Cannot convert existing claim value of type '%s' to desired type " +
+            "'%s'. JJWT only converts simple String, Date, Long, Integer, Short and Byte types automatically. " +
+            "Anything more complex is expected to be already converted to your desired type by the JSON Deserializer " +
+            "implementation. You may specify a custom Deserializer for a JwtParser with the desired conversion " +
+            "configuration via the JwtParserBuilder.deserializeJsonWith() method. " +
+            "See https://github.com/jwtk/jjwt#custom-json-processor for more information.";
```
lhazlewood: I think we should add one more sentence at the end of this paragraph:

"If using Jackson, you can specify custom claim POJO types as described in <url-here>."

Where <url-here> is of course the section in the documentation that explains how to do this that you're presumably writing in #500 

Since I would think Jackson will be the most widely used parser.

Thoughts?

extensions/jackson/src/main/java/io/jsonwebtoken/jackson/io/JacksonDeserializer.java
```diff
@@ -35,6 +42,32 @@ public JacksonDeserializer() {
         this(JacksonSerializer.DEFAULT_OBJECT_MAPPER);
     }
 
+    /**
+     * Creates a new JacksonDeserializer where the values of the claims can be parsed into given types. A common usage
```
lhazlewood: Can we add some documentation here that it uses a new `ObjectMapper` instance?

For example, after the current documentation:

```
<p>
Because custom type parsing requires modifying Jackson {@code ObjectMapper} state, this 
constructor creates a new internal {@code ObjectMapper} instance and customizes it to support the 
specified {@code claimTypeMap}.  This ensures that JJWT parsing behavior does not unexpectedly 
modify the state of another application-specific {@code ObjectMapper}.
</p>
<p>
If you would like to use your own {@code ObjectMapper} instance that also supports custom types for 
JWT {@code Claims}, you will need to first customize your {@code ObjectMapper} instance by registering 
your custom types and then use the {@link JacksonDeserializer(ObjectMapper) } constructor instead.
</p>
```

Based on this, I wonder if we might want to make `MappedTypeDeserializer` a dedicated independent public class in the `jjwt-jackson` module so that users can use it on their own ObjectMapper instance (e.g. they might have customized pretty printing, but still want custom type registration, etc).  That class would document how to use it in the class JavaDoc based on what we've done, e.g.:

```java
MappedTypeDeserializer typeDeserializer = new MappedTypeDeserializer(typeMap);
SimpleModule module = new SimpleModule();
module.addDeserializer(Object.class, typeDeserializer);
objectMapper.registerModule(module);
```

All of this would help address the confusion I had (for example) around type customization on an application-wide deserializer, as well as provide an out-of-the-box solution for people that wanted to customize their own ObjectMapper.

Thoughts?

bdemers: Great doc suggestion

I thought about extracting `MappedTypeDeserializer` out, but I worry it might cause a maintenance headache trying to cover more parsing edge cases (when combined with other Jackson configurations).  My personal pref would be to hold off unless people ask for it (YAGNI?).  It's really easy to extract that class later, but if we make invalid assumptions now, it is harder to retract them.

That was my thinking anyway, though I could be missing some of the benefits of sharing an ObjectMapper instance?

lhazlewood: > I worry it might cause a maintenance headache trying to cover more parsing edge cases (when combined with other Jackson configurations)

I was thinking the current implementation would be the only one we'd support, and if that doesn't meet one's needs, they can just do it the manual approach or implement their own `JsonSerializer`.  At least they'd see easily how to do it themselves since understanding how that works is most of the difficulty, and our example gets them out of the weeds.

That said, I see value with leaving it private and the YAGNI argument.  I guess I'm on the fence 50/50 and would be happy with whatever you choose, so feel free ;)

bdemers:
Rebased manually (to merge in the split package change)

Issue #388, opened 2018-08-15
java.lang.NoClassDefFoundError: org.json.JSONString


ITWall:
When I use 0.10.4 version, I got this error when I do the sample in readme. I don't know why but after add this library http://mvnrepository.com/artifact/org.json/json, I solved it. Not so sure if this is a bug or not.

lhazlewood:
Sigh.  Android is killing me! 😭 

org.json.JSONString and org.json.JSONWriter are used by the implementation but never used by Android at runtime.  When the class is loaded, that's when it barfs.  I'll have to change the implementation again, potentially duplicating logic from later org.json releases that Android doesn't have.

Thanks for reporting this.  You're correct - the workaround in the meantime is to include the org.json .jar.

rodmaz:
Same problem here on Android. Has this fixed on `0.10.5`? 
When we include version `0.10.5` we get `Gradle sync failed` on Gradle/Android Studio.

rodmaz:
This was causing the reported `freeze` when we called the `compact()` method on Android.

lhazlewood:
0.10.5 is being released as we speak.  Maven Central staging is *super* slow due to degraded network performance, so please allow it 30 minutes or so before trying.

rodmaz:
Thanks. We tried it like 1h ago as the tag was already there. We will resume tomorrow. Thanks again @lhazlewood!

fernando-jascovich:
Hi, I think that this issue (or some new form of it) is still around; when I set dependencies as explained in documentation: 
```
    runtimeOnly('io.jsonwebtoken:jjwt-orgjson:0.10.5') {
        exclude group: 'org.json', module: 'json' //provided by Android natively
    }
```
And then, I use the following example code:
```
        val token = Jwts.builder()
                .claim("myclaim", "myclaimValue")
                .signWith(SignatureAlgorithm.HS256, acceptableKey.toByteArray())
                .compact()
```
I get the following NPE:
```
java.lang.IllegalStateException: Unable to serialize header to json.
	at io.jsonwebtoken.impl.DefaultJwtBuilder.base64UrlEncode(DefaultJwtBuilder.java:381)
	at io.jsonwebtoken.impl.DefaultJwtBuilder.compact(DefaultJwtBuilder.java:333)
...
Caused by: io.jsonwebtoken.io.SerializationException: Unable to serialize object of type io.jsonwebtoken.impl.DefaultJwsHeader to JSON: null
[...]

PR #496, opened 2019-09-19

bdemers:
@jaapcoomans I'm dusting this off, sorry for the delay #458

I love the `FakeServiceDescriptorClassLoader` trick!

TODO:
- [ ] ~~Deprecate the `Classes` util class? (not sure if this is possible given the last two remaining usages of this class:~~
  - OrgJsonSerializer re: #388
  - RuntimeEnvironment (checking if Bouncy Castle is available)
- [x] merge in pending changes of #488 (rename Serializer classes)

IMHO, that first item does NOT block this issue, but I'd like to hear if anyone has any thoughts on the topic.


impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
```diff
@@ -296,9 +296,8 @@ public String compact() {
 
         if (this.serializer == null) {
             //try to find one based on the runtime environment:
-            InstanceLocator<Serializer<Map<String,?>>> locator =
```
bdemers: @lhazlewood Any idea why this was using reflection for a class in the same module?  Maybe an artifact of some older refactor?

lhazlewood: I believe this was old cruft that wasn't cleaned up when we made the `api` and `impl` modules.  The new change looks correct.

api/src/main/java/io/jsonwebtoken/lang/ImplementationNotFoundException.java
```diff
@@ -0,0 +1,20 @@
+package io.jsonwebtoken.lang;
+
+import io.jsonwebtoken.JwtException;
+
+/**
+ * Exception indicating that no implementation of an jjwt-api SPI was found on the classpath.
+ */
+public class ImplementationNotFoundException extends JwtException {
```
lhazlewood: I don't know why, but I've never liked the JDK classes that are like FooNotFoundException or WhateverNotBlahException - I've always felt like they didn't flow well and were clunky for the English language.

Can we make this `UnavailableImplementationException`?

Also, should this extend from some other exception instead of `JwtException` directly?

api/src/main/java/io/jsonwebtoken/lang/ImplementationNotFoundException.java
```diff
@@ -0,0 +1,20 @@
...
+
+    private static final String DEFAULT_NOT_FOUND_MESSAGE = "No implementation of %s found on the classpath. Make sure to include an implementation of jjwt-api.";
```
lhazlewood: Perhaps:

`"Unable to find an implementation for %s using java.util.ServiceLoader.  Ensure you include a backing implementation .jar in the classpath, for example jjwt-impl.jar, or your own .jar for custom implementations."`

api/src/main/java/io/jsonwebtoken/lang/Services.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Loads and instantiates all service implementation of the given SPI class and returns them as a List.
+     *
+     * @param spi The class of the Service Provider Interface
+     * @param <T> The type of the SPI
+     * @return An unmodifiable list with an instance of all available implementations of the SPI. No guarantee is given
+     * on the order of implementations, if more than one.
+     */
+    public static <T> List<T> loadAllAvailableImplementations(Class<T> spi) {
```
lhazlewood: This method name feels a little verbose to me.  How about just `loadAll` since this better parallels the `loadFirst` nomenclature below (i.e. it's not `loadFirstAvailableImplementation`, it's just a nice clean `loadFirst`).

api/src/main/java/io/jsonwebtoken/Jwts.java
```diff
@@ -15,19 +15,19 @@
...
  * Factory class useful for creating instances of JWT interfaces.  Using this factory class can be a good
  * alternative to tightly coupling your code to implementation classes.
  *
- * @since 0.1
+ * @since 0.11
  */
 public final class Jwts {
 
-    private static final Class[] MAP_ARG = new Class[]{Map.class};
+    private static final JwtFactory FACTORY = Services.loadFirst(JwtFactory.class);
```
lhazlewood: See note above about retaining the current implementation.

api/src/main/java/io/jsonwebtoken/lang/Services.java
lhazlewood: Also, is this method called anywhere else other than by `loadFirst` ?  If not, we should make it private to reduce our exposure area of things we need to support for users.

api/src/main/java/io/jsonwebtoken/security/KeyGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+import io.jsonwebtoken.SignatureAlgorithm;
+
+import javax.crypto.SecretKey;
+
+/**
+ * Implementations of KeyGenerator will generate {@link SecretKey} based for a given {@link SignatureAlgorithm}.
+ *
+ * @since 0.11.0
+ */
+public interface KeyGenerator {
```
lhazlewood: I'm not sure about this interface or `KeyPairGenerator`.  In the very new future (1.0), SignatureAlgorithm will be an interface, not an enum, to allow for custom signature algorithm implementations.

Based on this, I'm wondering if the new `SignatureAlgorithm` implementation should implement new interfaces, `SecretKeyAlgorithm` and `KeyPairAlgorithm` and those interfaces have `generateKey` and `generateKeyPair` methods respectively.

Anyway, is there any harm in retaining the current runtime classpath lookup behavior until we can migrate the enums to an interface?

api/src/main/java/io/jsonwebtoken/security/KeyPairGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+import io.jsonwebtoken.SignatureAlgorithm;
+
+import java.security.KeyPair;
+
+/**
+ * Implementations of KeyGenerator will generate {@link KeyPair} based for a given {@link SignatureAlgorithm}.
+ *
+ * @since 0.11.0
+ */
+public interface KeyPairGenerator {
```
lhazlewood: See `KeyGenerator` comments.

bdemers:
@lhazlewood I've reduced the scope of this PR down to the `CompressionCodecs` and `Serializer`/`Deserializer` implementations.
We can take another look at the other changes after we make any breaking changes for 1.0

bdemers:
Good call on the `DefaultCompressionCodecResolver`.  Ensuring JJWT's codecs are used first gets a little sticky.  We _could_ add a package name check on the resolved Codec, but at that point, we could just use reflection to load the implementation (without the Service Loader)

I like the `.compressWith(name)` idea! The previous concerns about ensuring JJWT's codecs are used, makes me wonder if we should just worry about the Serializer/Deserializer for this release? (and pull the CompressionCodec out of this PR), I think i'm on the fence now.

?

lhazlewood:
hrm - you've gone through all this work, it'd feel like taking an unnecessary step back IMO if we don't just finish these few extra things.

Ensuring JJWT's codecs are used first is pretty easy IMO, unless I'm missing something:

In `DefaultCompressionCodecResolver`, after line 62 but before line 64, you call `CompressionCodecs.forName`.  If still null after that, the existing line 64 throws the exception as expected.  No?  What am I missing? 😅

lhazlewood:
In fact, move lines 57-62 into `CompressionCodecs.forName` and both the resolver and the parser builder call that method.  The JJWT-prioritization logic is localized into one place and works in both scenarios (AFAICT).

bdemers:
`DefaultCompressionCodecResolver` (or anything in the impl module) isn't an issue.  I was thinking about how ensuring the JJWT CompressionCodec implementations would affect the proposed changes to [CompressionCodecs](https://github.com/jwtk/jjwt/pull/496/files#diff-8ed21ed2008456dd9fba70aa63655cd6)

`CompressionCodec` would need to revert back to using reflection to make sure `CompressionCodecs.DEFLATE` and `DefaultCompressionCodecResolver.resolveCompressionCodec({"zip": "deflate"})` both return the same implementation/instance.

Since the library, itself wouldn't be taking advantage of loading it's CompressionCodec's this way, I was questioning if we should continue down this path, and whether this change would be making this extensible just for the sake of extensibility. (**NOTE:** I have zero data on how often custom implementations are used `<insert grain of salt>`)

Either way, adding an additional method to `CompressionCodecResolver` might be handy:

```java
CompressionCodec resolveCompressionCodec(String name) throws CompressionException;
```

This could be used for your proposed `. compressWith(name)` method (service-loader or not).

Thinking out loud.

lhazlewood:
I'm suggesting to keep the logic that you've added (public static final variables with static class init block).  `CompressionCodecs.GZIP` and `DEFLATE` need to remain for backwards compatibility.

I'm proposing this (simplified for brevity - needs javadoc comments, etc):

```java
private static final String DEFLATE_IMPL_CLASSNAME = "io.jsonwebtoken.impl.compression.DeflateCompressionCodec";
private static final String GZIP_IMPL_CLASSNAME = "io.jsonwebtoken.impl.compression.GzipCompressionCodec";

private static final Map<String,CompressionCodec> CODECS;
public static final CompressionCodec DEFLATE;
public static final CompressionCodec GZIP;

static {
    CODECS = new HashMap<String,CompressionCodec>();
    Collection<CompressionCodec> loaded = Services.loadAll(CompressionCodec.class);
    for (CompressionCodec codec : loaded) {
        String className = codec.getClass().getName();
        if (DEFLATE_IMPL_CLASSNAME.equals(className)) {
          DEFLATE = codec;
        } else if (GZIP_IMPL_CLASSNAME.equals(className)) {
          GZIP=codec;
        }
        String name = codec.getName();
        // TODO: if !Strings.hasText(name) throw exception
       name = name.toUpperCase() //guarantee canonical for map storage and lookup
       CODECS.put(name, codec);
    }

    //TODO: if either DEFLATE or GZIP are null here, throw an exception
    
    // Now ensure JJWT's GZIP and DEFLATE always take precedence for RFC-standard names:
    CODECS.put(DEFLATE.getName(), DEFLATE);
    CODECS.put(GZIP.getName(), GZIP);
}

public static final CompressionCodec forName(String name) {
  if (!Strings.hasText(name)) { ... throw exception ... }
  String name = name.toUpperCase(); //normalize for map lookup
  CompressionCodec codec = CODECS.get(name);
  if (codec == null) { throw new UnsupportedCodecException("There is no codec with algorithm name " + name + ": Ensure jjwt-impl.jar or your custom implementation .jar are in the classpath according to the documentation at https://github.com/jwtk/jjwt#whatever." }
}
```

lhazlewood:
Hrm - the reflection angle is super easy too.  Maybe a hybrid approach makes sense:  do reflection for our internal implementations (no serviceloader metadata) and allow custom CompressionCodecs to be added to the classpath for parser lookup?

I like the reflection angle for our internal stuff actually, and serviceloader for plugins.  I think that's super clean as we wouldn't conflict with any plugins, ever.

lhazlewood:
So based on that last comment, how about this:

```java
private static final String DEFLATE_IMPL_CLASSNAME = "io.jsonwebtoken.impl.compression.DeflateCompressionCodec";
private static final String GZIP_IMPL_CLASSNAME = "io.jsonwebtoken.impl.compression.GzipCompressionCodec";

private static final Map<String,CompressionCodec> CODECS;
public static final CompressionCodec DEFLATE;
public static final CompressionCodec GZIP;

static {
    CODECS = new HashMap<String,CompressionCodec>();
    Collection<CompressionCodec> loaded = Services.loadAll(CompressionCodec.class);
    for (CompressionCodec codec : loaded) {
        String name = codec.getName();
        // TODO: if !Strings.hasText(name) throw exception
       name = name.toUpperCase() //guarantee canonical for map storage and lookup
       CODECS.put(name, codec);
    }

    DEFLATE = Classes.newInstance(DEFLATE_IMPL_CLASSNAME);
    CODECS.put(DEFLATE.getName, DEFLATE);

    GZIP = Classes.newInstance(GZIP_IMPL_CLASSNAME);
    CODECS.put(GZIP.getName(), GZIP);
}

public static final CompressionCodec forName(String name) {
  if (!Strings.hasText(name)) { ... throw exception ... }
  String name = name.toUpperCase(); //normalize for map lookup
  CompressionCodec codec = CODECS.get(name);
  if (codec == null) { throw new UnsupportedCodecException("There is no codec with algorithm name " + name + ": Ensure jjwt-impl.jar or your custom implementation .jar are in the classpath according to the documentation at https://github.com/jwtk/jjwt#whatever." }
}
```

bdemers:
@lhazlewood getting back to this.
I did something very similar to your suggestion, except I moved the logic into `DefaultCompressionCodecResolver`

impl/src/test/groovy/io/jsonwebtoken/impl/io/RuntimeClasspathSerializerLocatorTest.groovy
```diff
@@ -98,36 +109,27 @@ class RuntimeClasspathSerializerLocatorTest {
...
-            }
-        }
+        prepareFakeServiceClassLoader(TEST_SERVICE_DESCRIPTOR_GSON)
 
-        def serializer = locator.getInstance()
+        def serializer = new RuntimeClasspathSerializerLocator().getInstance()
         assertTrue serializer instanceof GsonSerializer
     }
+
+    private void prepareNoServiceDescriptorClassLoader() {
```
bdemers: doh, i missed this class, i need to clean this test up a little to make it look like the Deserializer equivalent

impl/src/main/resources/META-INF/services/io.jsonwebtoken.JwtFactory
```diff
@@ -0,0 +1 @@
+io.jsonwebtoken.impl.DefaultJwtFactory
```
bdemers: TODO: delete

bdemers:
a couple more minor tweaks found in a quick self review

impl/src/main/java/io/jsonwebtoken/impl/compression/DefaultCompressionCodecResolver.java
```diff
@@ -54,19 +76,23 @@ public CompressionCodec resolveCompressionCodec(Header header) {
...
     }
 
     private String getAlgorithmFromHeader(Header header) {
         Assert.notNull(header, "header cannot be null.");
 
         return header.getCompressionAlgorithm();
     }
+
+    private CompressionCodec byName(String name) {
+        Assert.hasText("'name' must not be empty");
```
lhazlewood: Looks like an invalid assertion - not passing in the `name` argument 😉

impl/src/main/java/io/jsonwebtoken/impl/compression/DefaultCompressionCodecResolver.java
```diff
@@ -54,19 +76,23 @@ public CompressionCodec resolveCompressionCodec(Header header) {
...
+
+        CompressionCodec codec = codecs.get(name.toUpperCase());
+        if (codec == null) {
+            throw new CompressionException(String.format(MISSING_COMPRESSION_MESSAGE, name));
```
lhazlewood: I don't know if it's valid to use a CompressionException here, is it?  I mean, no compression/decompression is occurring - it's just a lookup failure.  Should this be something else?

api/src/main/java/io/jsonwebtoken/lang/Services.java
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * @param spi The class of the Service Provider Interface
+     * @param <T> The type of the SPI
+     * @return An unmodifiable list with an instance of all available implementations of the SPI. No guarantee is given
+     * on the order of implementations, if more than one.
+     */
+    public static <T> List<T> loadAll(Class<T> spi) {
+        ServiceLoader<T> serviceLoader = ServiceLoader.load(spi);
+
+        List<T> implementations = new ArrayList<>();
```
lhazlewood: if `implementations` is empty,  this method will eventually return an empty (immutable) collection.  We should throw an `UnavailableImplementationException` instead to indicate an invalid state or spi class.

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java
```diff
@@ -45,12 +46,10 @@
 
     @SuppressWarnings("WeakerAccess") //to allow testing override
     protected Deserializer<T> locate() {
-        if (isAvailable("io.jsonwebtoken.jackson.io.JacksonDeserializer")) {
-            return Classes.newInstance("io.jsonwebtoken.jackson.io.JacksonDeserializer");
-        } else if (isAvailable("io.jsonwebtoken.orgjson.io.OrgJsonDeserializer")) {
-            return Classes.newInstance("io.jsonwebtoken.orgjson.io.OrgJsonDeserializer");
-        } else if (isAvailable("io.jsonwebtoken.gson.io.GsonDeserializer")) {
-            return Classes.newInstance("io.jsonwebtoken.gson.io.GsonDeserializer");
+        ServiceLoader<Deserializer> serviceLoader = ServiceLoader.load(Deserializer.class);
```
lhazlewood: Can this entire class be replaced by a call to `Services.loadFirst` ?

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathSerializerLocator.java
```diff
@@ -17,21 +17,22 @@
...
 
+import java.util.Iterator;
+import java.util.ServiceLoader;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * @since 0.10.0
  */
-public class RuntimeClasspathSerializerLocator implements InstanceLocator<Serializer> {
+public class RuntimeClasspathSerializerLocator<T> implements InstanceLocator<Serializer<T>> {
```
lhazlewood: Can this entire class be replaced by a call to `Services.loadFirst` ?

impl/src/main/java/io/jsonwebtoken/impl/compression/DefaultCompressionCodecResolver.java
bdemers: 🤦‍♂

impl/src/main/java/io/jsonwebtoken/impl/compression/DefaultCompressionCodecResolver.java
bdemers: I had mixed feelings about that one too, in the end, I just kept the functionality the same (the previous version would throw a `CompressionException` if the codec could not be found.

impl/src/main/java/io/jsonwebtoken/impl/compression/DefaultCompressionCodecResolver.java
bdemers: (also, good catch)

skjolber:
So will this all work with JDK9+ modules now?

impl/src/main/java/io/jsonwebtoken/impl/io/RuntimeClasspathDeserializerLocator.java
bdemers: Yes! yes it can

bdemers:
@skjolber the big blocker for 9+ was the split package issue in:  #488 
(We are testing with Open JDK 8, 9, 10, 11 now too)

If you run into anything specific let us know!

We haven't added module-info's as we are still technically testing/building on 1.7
There is another issue covering this specifically, but we _could_ add them starting with the 1.0 release, but it would mean that we would need to use Java 11 for building (and still target 1.8)
i.e. this: https://maven.apache.org/plugins/maven-compiler-plugin/examples/module-info.html

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
```diff
@@ -295,10 +294,8 @@ public JwtBuilder claim(String name, Object value) {
     public String compact() {
 
         if (this.serializer == null) {
-            //try to find one based on the runtime environment:
-            InstanceLocator<Serializer<Map<String,?>>> locator =
-                Classes.newInstance("io.jsonwebtoken.impl.io.RuntimeClasspathSerializerLocator");
-            this.serializer = locator.getInstance();
+            // try to find one based on the services available
+            this.serializer = Services.loadFirst(Serializer.class);
```
lhazlewood: The only concern I have about this is that the runtime exception type will have changed since the previous implementation, and if anyone is catching that, their current code will not function correctly.  Perhaps we should wrap this exception with the previous one to guarantee compatibility?

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -255,12 +247,11 @@ public boolean isSigned(String jwt) {
...
-        // TODO move this to constructor before 1.0
+        // TODO, this logic is only need for a now deprecated code path
+        // remove this block in v1.0 (the equivalent is already in DefaultJwtParserBuilder)
         if (this.deserializer == null) {
-            //try to find one based on the runtime environment:
-            InstanceLocator<Deserializer<Map<String, ?>>> locator =
-                Classes.newInstance("io.jsonwebtoken.impl.io.RuntimeClasspathDeserializerLocator");
-            this.deserializer = locator.getInstance();
+            // try to find one based on the services available
+            this.deserializer = Services.loadFirst(Deserializer.class);
```
lhazlewood: same comment w.r.t. runtime exception type differences

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParserBuilder.java
```diff
@@ -170,6 +172,14 @@ public JwtParserBuilder setCompressionCodecResolver(CompressionCodecResolver com
 
     @Override
     public JwtParser build() {
+
+        // Only lookup the deserializer IF it is null. It is possible a Deserializer implementation was set
+        // that is NOT exposed as a service and no other implementations are available for lookup.
+        if (this.deserializer == null) {
+            // try to find one based on the services available:
+            this.deserializer = Services.loadFirst(Deserializer.class);
```
lhazlewood: same

impl/src/test/groovy/io/jsonwebtoken/DeprecatedJwtsTest.groovy
```diff
@@ -43,7 +44,7 @@ class DeprecatedJwtsTest {
     }
 
     protected static String toJson(o) {
-        def serializer = new RuntimeClasspathSerializerLocator().getInstance()
+        def serializer = new Services().loadFirst(Serializer)
```
lhazlewood: Just curious - why are you instantiating this class?

Unless I'm missing something, it should be marked as static final and the constructor should be private.

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
lhazlewood: P.S. with a TODO statement to remove the wrapping for 1.0

impl/src/test/groovy/io/jsonwebtoken/DeprecatedJwtsTest.groovy
bdemers: Ahh, we didn't have a private constructor on `Services`, I'll add one, and fix this copypasta

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParserBuilder.java
bdemers: This one is a new interface, so we don't need to retain a previous behavior

api/src/main/java/io/jsonwebtoken/lang/Services.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+import java.util.ServiceLoader;
+
+/**
+ * Helper class for loading services from the classpath, using a {@link ServiceLoader}. Decouples loading logic for
+ * better separation of concerns and testability.
+ */
+public final class Services {
+
+    private Services() {}
```
bdemers: Added private constructor

api/src/main/java/io/jsonwebtoken/lang/UnknownClassException.java
```diff
@@ -48,16 +48,17 @@ public UnknownClassException(String message) {
...
     }
+    */
 
     /**
      * Constructs a new UnknownClassException.
      *
      * @param message the reason for the exception
      * @param cause   the underlying Throwable that caused this exception to be thrown.
-     *
+     */
```
bdemers: This diff looks funny, but this constructor was commented out, so I just uncommented it (and added a TODO to remove in the next version)

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -107,13 +108,6 @@ public DefaultJwtParser() { }
         this.base64UrlDecoder = base64UrlDecoder;
         this.deserializer = deserializer;
         this.compressionCodecResolver = compressionCodecResolver;
-
-        if (this.deserializer == null) {
-            //try to find one based on the runtime environment:
-            InstanceLocator<Deserializer<Map<String, ?>>> locator =
-                Classes.newInstance("io.jsonwebtoken.impl.io.RuntimeClasspathDeserializerLocator");
-            this.deserializer = locator.getInstance();
-        }
```
bdemers: NOTE: this constructor was after the last release, so moving this block to the DefaultJwtParser builder is _more_ correct

api/src/main/java/io/jsonwebtoken/lang/UnavailableImplementationException.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Exception indicating that no implementation of an jjwt-api SPI was found on the classpath.
+ */
+public class UnavailableImplementationException extends RuntimeException {
+
+    private static final String DEFAULT_NOT_FOUND_MESSAGE = "Unable to find an implementation for %s using java.util.ServiceLoader. Ensure you include a backing implementation .jar in the classpath, for example jjwt-impl.jar, or your own .jar for custom implementations.";
+
+    public UnavailableImplementationException(final Class klass) {
+        super(String.format(DEFAULT_NOT_FOUND_MESSAGE, klass));
```
lhazlewood: This might be pedantic, but should we have a `notNull` assertion on `klass`?  Feel free to add it or not - just an observation.

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
lhazlewood: Seeing this elucidates something that the ParserBuilder is going to fix for applications:

In existing releases (before 0.11.0), if a `Deserializer` implementation wasn't found, the exception would be thrown when `parse` is called.  That means this exception wouldn't be discovered until well after the application has started - e.g. during an HTTP request.  Since the default Deserializer implementation lookup isn't based on runtime (e.g. request) input and instead based on the runtime classpath, it'd be much better for developers to know about this problem at application start time, not later (i.e. 'fail fast').

The new Parser Builder implementation can ensure that's the case because this logic is now in the `build` implementation - likely something to be called during application startup to create an application singleton (e.g. in Spring).

Much nicer!

api/src/main/java/io/jsonwebtoken/lang/Services.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Loads and instantiates all service implementation of the given SPI class and returns them as a List.
+     *
+     * @param spi The class of the Service Provider Interface
+     * @param <T> The type of the SPI
+     * @return An unmodifiable list with an instance of all available implementations of the SPI. No guarantee is given
+     * on the order of implementations, if more than one.
+     */
+    public static <T> List<T> loadAll(Class<T> spi) {
+        ServiceLoader<T> serviceLoader = ServiceLoader.load(spi);
```
lhazlewood: Since this is a public API class, I think we should have a `notNull` assertion on `spi`.

api/src/main/java/io/jsonwebtoken/lang/Services.java
```diff
@@ -0,0 +1,74 @@
...
+     * to find implementations. When multiple implementations are available it will return the first one that it
+     * encounters. There is no guarantee with regard to ordering.
+     *
+     * @param spi The class of the Service Provider Interface
+     * @param <T> The type of the SPI
+     * @return A new instance of the service.
+     * @throws UnavailableImplementationException When no implementation the SPI is available on the classpath.
+     */
+    public static <T> T loadFirst(Class<T> spi) {
+        ServiceLoader<T> serviceLoader = ServiceLoader.load(spi);
```
lhazlewood: Since this is a public API class, I think we should have a `notNull` assertion on `spi`.

api/src/main/java/io/jsonwebtoken/lang/UnavailableImplementationException.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.jsonwebtoken.lang;
+
+/**
+ * Exception indicating that no implementation of an jjwt-api SPI was found on the classpath.
+ */
+public class UnavailableImplementationException extends RuntimeException {
```
lhazlewood: Since this is a brand new public API class - we should probably lock this down and make it `final` (open/closed principle, etc).  We can always remove the modifier later if we need to subclass it for our own needs.

bdemers: 💯

api/src/main/java/io/jsonwebtoken/lang/UnavailableImplementationException.java
bdemers: I think `null` is a valid case here for a couple of reasons:
1. I'd rather not throw an exception from another exception
2. the assertion message will contain `Unable to find an implementation for %s` which while not extremely useful, is valid.  I think methods that throw this exception should be checking for null implementations.

That said, making the class final, and making this constructor package-private should reduce the possibility of this coming up.

Thoughts?

api/src/main/java/io/jsonwebtoken/lang/Services.java
bdemers: 💯

impl/src/main/java/io/jsonwebtoken/impl/BackwardCompatibilityUtil.java
```diff
@@ -0,0 +1,26 @@
...
+import io.jsonwebtoken.lang.Services;
+import io.jsonwebtoken.lang.UnavailableImplementationException;
+import io.jsonwebtoken.lang.UnknownClassException;
+
+/**
+ * A package private utility class to help retain backward compatibility, before breaking changes are made for v1.0.
+ * TODO: remove this class before v1.0
+ */
+@Deprecated
+final class BackwardCompatibilityUtil {
```
bdemers: This class name is ugly, but I don't have any better ideas.  Moved to a util class to make testing (coverage) easier

api/src/main/java/io/jsonwebtoken/lang/UnavailableImplementationException.java
lhazlewood: Ah, right - throwing an exception from an exception.  I guess I was in 'no nulls everywhere possible' mode.

I can see both sides - leaving it alone and still including the null assertion.  I'm on the fence but I'd probably pick putting a null assertion in there because it means that the class is being used illegally - the upstream code didn't check for null and that would surface a programming error/bug on our part, which is a good thing I suppose.

That said, have you verified that all upstream callers have checked for null first?  If so, I'm fine with whatever you decide.

impl/src/main/java/io/jsonwebtoken/impl/BackwardCompatibilityUtil.java
lhazlewood: How about `LegacyServices.loadFirst` ? And put it in the `io.jsonwebtoken.impl.lang` package to mirror the current `Services` class?  This way also ensures the method signature is the same, so when refactoring comes, we just remove the `Legacy` prefix on the respective line of code.

bdemers: Much better name! And I'm also less afraid to stick it in `.lang` with a name like that too

api/src/main/java/io/jsonwebtoken/lang/Services.java
```diff
@@ -0,0 +1,98 @@
...
+     */
+    @Deprecated
+    public static final class Legacy {
+
+        /**
+         * Wraps {@code Services.loadFirst} and throws a {@link UnknownClassException} instead of a
+         * {@link UnavailableImplementationException} to retain the previous behavior. This method should be used when
+         * to retain the previous behavior of methods that throw an unchecked UnknownClassException.
+         */
+        public static <T> T loadFirst(Class<T> spi) {
```
bdemers: @lhazlewood What do you think of `Services.Legacy` ?

lhazlewood: My preference is still `LegacyServices` in the `impl` module - I wouldn't want a new inner class that we're going to soon delete to show up in the public API if we can avoid it.

bdemers: OH, sorry I miss read your comment about `.impl.lang` vs `.lang` I'm on it

api/src/main/java/io/jsonwebtoken/lang/UnavailableImplementationException.java
lhazlewood: Just want to bump in case you missed this thread - If not putting in a null assertion, have you been able to verify upstream callers do so?

bdemers:
All refs to `new UnavailableImplementationException()`  are from `Servies.*` and all refs to those method pass a `Class` directly in.

PR #498, opened 2019-09-24
Updating to jackson-databind 2.9.10

bdemers:


lhazlewood:
You have to change the docs to reflect this as well.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## Release Notes
 
+### 0.11.0
+
+* Updates the Jackson dependency version to [2.9.10](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9#patches)
+to address three security vulnerabilities in Jackson:
```
lhazlewood: Oh, I was referring to the actual user doc (Readme) - we reference the jackson version in there I think.

bdemers: I didn't see one with a quick grep

lhazlewood: Ah, right, I was thinking of the dependency declaration part of the docs, but that's JJWT's .jar, not Jackson's.

lhazlewood:
@bdemers you didn't squash :(

bdemers:
@lhazlewood 
I used the squash button in GitHub, it shows up in the PR as the original commits, but master has the squashed version: https://github.com/jwtk/jjwt/commits/master

lhazlewood:
Iiiiiinteresting - I didn't know!  Thanks for educating me! 😄

bdemers:
We _could_ change the merge policy to _only_ allow the squash button.
image

I have mixed feelings about that, because there are some cases where a merge commit _might_ be a better choice (but there are ways to work around that too).

Anyway, I figured I'd throw out the idea

lhazlewood:
Let's try it with only squash and rebase merging. If we don't like it, we can easily revert.  Also I would definitely want to delete head branches on merge - that's a super convenient option.

diego-ruiz-rei:
hello! Is there going to be a release soon that includes these updates?

bdemers:
That is the plan! but until we get the release out you may want to manage the version of Jackson directly.

PR #500, opened 2019-09-24
Docs: Adds section to README covering custom object parsing

bdemers:
Doc: #495

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ enforcement.
 * [JSON Processor](#json)
   * [Custom JSON Processor](#json-custom)
   * [Jackson ObjectMapper](#json-jackson)
-    * [Custom Claim Types](#json-jackson-custom-parsing)
+    * [Custom Claim Types](#json-jackson-custom-types)
 * [Base64 Codec](#base64)
   * [Custom Base64 Codec](#base64-custom)
 
@@ -1311,7 +1311,7 @@ Jwts.parser()
     // ... etc ...
 ```
 
-<a name="json-jackson-custom-parsing"></a>
+<a name="json-jackson-custom-types"></a>
 ### Parsing of Custom Claim Types
 
 By default JJWT will only convert simple claim types: String, Date, Long, Integer, Short and Byte.  If you need to deserialize other types you can configure the `JacksonDeserializer` by passing a `Map` of claim names to types in through a constructor. For example:
@@ -1344,7 +1344,7 @@ Jwts.parserBuilder()
     .build()
     .getBody()
     
-    get("user", User.class)
+    .get("user", User.class)
 ```
 
 <a name="base64"></a>
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ enforcement.
 * [JSON Processor](#json)
   * [Custom JSON Processor](#json-custom)
   * [Jackson ObjectMapper](#json-jackson)
+    * [Custom Claim Types](#json-jackson-custom-parsing)
 * [Base64 Codec](#base64)
   * [Custom Base64 Codec](#base64-custom)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1310,6 +1311,42 @@ Jwts.parser()
...
+Jwts.parserBuilder()
+
+    .deserializeJsonWith(new JacksonDeserializer(Maps.of("user", User.class)))
+
+    .build()
+    .parseClaimsJwt(aJwtString)
+    .build()
+    .getBody()
+    
+    get("user", User.class)
```
lhazlewood: Should have a dot prefix (`.get(...`)

README.md
```diff
@@ -1310,6 +1311,42 @@ Jwts.parser()
...
+```java
+Jwts.parserBuilder()
+
+    .deserializeJsonWith(new JacksonDeserializer(Maps.of("user", User.class)))
+
+    .build()
+    .parseClaimsJwt(aJwtString)
+    .build()
+    .getBody()
+
```
lhazlewood: Blank?

README.md
```diff
@@ -1310,6 +1311,42 @@ Jwts.parser()
     // ... etc ...
 ```
 
+<a name="json-jackson-custom-parsing"></a>
```
lhazlewood: Let's have this anchor be `json-jackson-custom-types` to match the section header and content (better SEO)

README.md
```diff
@@ -1310,6 +1311,42 @@ Jwts.parser()
...
+### Parsing of Custom Claim Types
+
+By default JJWT will only convert simple claim types: String, Date, Long, Integer, Short and Byte.  If you need to deserialize other types you can configure the `JacksonDeserializer` by passing a `Map` of claim names to types in through a constructor. For example:
+
+```java
+new JacksonDeserializer(Maps.of("user", User.class))
```
lhazlewood: Shouldn't we show the constructor with `ObjectMapper` as an argument?

README.md
```diff
@@ -1310,6 +1311,42 @@ Jwts.parser()
...
+      "lastName": "Coder"
+    }
+}
+```
+
+The `User` object could be retrieved from the `user` claim with the following code:
+
+```java
+Jwts.parserBuilder()
+
```
lhazlewood: Why the blank line?  I don't think we do this in other parts of the doc, do we?  Whatever we do, we should keep it consistent.

README.md
```diff
@@ -1310,6 +1311,42 @@ Jwts.parser()
...
+    .deserializeJsonWith(new JacksonDeserializer(Maps.of("user", User.class)))
+
```
lhazlewood: Blank?

README.md
bdemers: I was trying to follow what I thought was the existing convention, i.e.:
```java
Jwts.parser()

    .setSigningKeyResolver(signingKeyResolver) // <----
    
    .parseClaimsJws(jwsString);
```

My personal pref is would be to strip those empty lines

README.md
bdemers: This constructor will create and manipulate a new `ObjectMapper` (conventionally mapping the claims to types).  This could cause unforeseen issues with an existing object mapper.

Anything more _advanced_ then this, the `JacksonDeserializer(ObjectMapper)` constructor should be used (as shown in the previous section).

README.md
bdemers: I was just trying to follow what I thought was the existing format, i.e.:

```java
Jwts.parser()
    
    .setAllowedClockSkewSeconds(seconds) // <----
    
    // ... etc ...
    .parseClaimsJws(jwt);
```

My personal pref would be to strip the empty lines.  My assumption was probably wrong, maybe the convention is to insert a new line between each line of code? (and keep any relevant comments directly above a given line)

README.md
lhazlewood: Ah, I see what we did - it was to make it more visually obvious as to what is calling attention.  Let's stick to the current convention, and we can have another PR if we want to change that convention.

Now that I'm looking at it, the extra lines with the `// <----- ` makes it more visually obvious and probably enhances readability.  Just my .02 though.

README.md
lhazlewood: My point is shouldn't there be a `JacksonDeserializer(ObjectMapper, Map)` constructor?

README.md
lhazlewood: Ah, based on current convention, let's add a `// <----- ` comment at the end of line 1340

README.md
bdemers: No, I think if you want to use a custom ObjectMapper, you would be in charge of configuring that instance.  This constructor modifies the ObjectMapper, which could invalidate, or worse only slightly change how the ObjectMapper was originally configured.

You could end up in a situation where the same ObjectMapper is used for a few requests, then a JwtParser is constructed for the first time (with `JacksonDeserializer(ObjectMapper, Map)`) and this could change how future requests were parsed with the `ObjectMapper` in that application.  For example, I couldn't use `JacksonSerializer.DEFAULT_OBJECT_MAPPER` for the same reason, as it would have changed how other JacksonDeserializers would parse.

That said this issue is specific to the readme, lets me move this over to #495 (I'll come back and tweak the doc with the signatures that we agree with)
https://github.com/jwtk/jjwt/pull/495/files#diff-43ca7dcadb15963ec66e63df02544e1aR68

lhazlewood: Ah, ok, this is super helpful.

I think we need to be super clear about this in the documentation. For example, the very first sentence in the [first section on Jackson](https://github.com/jwtk/jjwt#jackson-json-processor) states:

> If you have an application-wide Jackson ObjectMapper (as is typically recommended for most applications), you can eliminate the overhead of JJWT constructing its own ObjectMapper by using yours instead.

I'm sure people would be curious why they can use their own ObjectMapper in one scenario, but not another, so this should be crystal clear in the docs IMO (i.e. in this PR).

bdemers:
GitHub - Travis sync error, the build above is green, it shows up yellow/pending here.
merging...

PR #501, opened 2019-09-24
Add support for testing with JDK11

bdemers:
There were a couple issues preventing running the build with Java 11
- A groovy bug - fixed with a version bump https://issues.apache.org/jira/browse/GROOVY-8727
- The ASM version that is embedded in easymock does NOT support Java 11 classes - worked around by stubbing out those usages instead
- javadoc modules issue - fixed by setting the Javadoc source version to 1.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ jdk:
   - oraclejdk8
   - oraclejdk9
   - openjdk10
+  - openjdk11
 
 before_install:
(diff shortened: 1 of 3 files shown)


bdemers:
**NOTE:** I'm not sure if `openjdk11` will work with `trusty`, or if we will need to use `oraclejdk11` instead, I've seen mixed reports on the web so I'll just try one and then the other.

bdemers:
@lhazlewood the Java 11 workaround isn't pretty, but it is contained to `SignatureAlgorithmTest.groovy`

PR #503, opened 2019-09-26
Reenable using the manifest from maven-bundle-plugin run

### LINKED ISSUE #502: The jars are not valid OSGi bundles anymore
Since this commit:
https://github.com/jwtk/jjwt/commit/439c027c2effcbc0e28d4507e8201071bd14f5d4

The jars do not contain the OSGi metadata anymore. I can do a PR to add it back.

cschneider:
Resolves #502

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -435,13 +435,7 @@
                 </executions>
                 <configuration>
                     <instructions>
-                        <Import-Package><![CDATA[
-                            android.util;resolution:=optional,
-                            org.bouncycastle.jce;resolution:=optional,
-                            org.bouncycastle.jce.spec;resolution:=optional,
-                            *
-                        ]]>
-                        </Import-Package>
+                        <_include>-bnd.bnd</_include>
                     </instructions>
                 </configuration>
             </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -304,6 +304,10 @@
                 <version>${maven.jar.version}</version>
                 <configuration>
                     <archive>
+                        <manifest>
+                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                        </manifest>
                         <manifestFile>
                             ${project.build.outputDirectory}/META-INF/MANIFEST.MF
                         </manifestFile>
--- a/pom.xml
+++ b/pom.xml
@@ -302,19 +302,13 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
                 <version>${maven.jar.version}</version>
-                <!--
                 <configuration>
                     <archive>
-                        <manifest>
-                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
-                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
-                        </manifest>
                         <manifestFile>
                             ${project.build.outputDirectory}/META-INF/MANIFEST.MF
                         </manifestFile>
                     </archive>
                 </configuration>
-                -->
             </plugin>
             <!-- Allow for writing tests in Groovy: -->
             <plugin>


lhazlewood:
Does this address #502 ?

pom.xml
```diff
@@ -302,19 +302,13 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
                 <version>${maven.jar.version}</version>
-                <!--
                 <configuration>
                     <archive>
-                        <manifest>
```
lhazlewood: Why were these two entries removed?

cschneider: I was not sure about those entries .. I removed them to not introduce more changes than necessary to the current state. As they have been there before they might be safe.

lhazlewood: I'm confused 😄 

How is OSGi supported per this PR?  I don't see the bundle plugin being used, so I'm not exactly sure what is going on here 😅 

Also, can you please re-enable these lines and see if things still work?  I'd like to retain that metadata if it doesn't hurt anything.

cschneider: I reenabled the two other settings now.
The OSGi headers are created in two steps:
1. The maven-bundle-plugin creates an enhanced Manifest file at https://github.com/jwtk/jjwt/blob/c246385be1b4beca4a5cd1a2debc1f5aa5008848/pom.xml#L432-L457
2. The jar plugin then must be configured to use this changed Manifest

The current state on master is that only step 1 happens.
So my PR is meant to reenable step 2 again.

You can check the effect by looking into the Manifest of the created jar files before and after the change.

lhazlewood: Ah, I forgot about the bundle plugin in the pom - thanks for clarifying.

That said, will this configuration work for every module now that we've gone to a multi-module project structure?

That is, we have an `api` module, an `impl` module, and a module for each of the JSON processor implementations (`jackson`, `orgjson`, `gson`).  Is the OSGi metadata valid for all of them?  Or does the bundle plugin configuration need to be re-defined for each module?

cschneider: Normally the default configuration works fine. I will give feedback if we spot issues.

lhazlewood: I meant the current configuration of the bundle plugin:

```
                <configuration>
                    <instructions>
                        <Import-Package><![CDATA[
                            android.util;resolution:=optional,
                            org.bouncycastle.jce;resolution:=optional,
                            org.bouncycastle.jce.spec;resolution:=optional,
                            *
                        ]]>
                        </Import-Package>
                    </instructions>
                </configuration>
```

Is this valid for every module in the project including the new module-per-json-processor approach?

cschneider: Ah .. got it. 
I think I can achieve the same without the config. 
If I make the dependencies optional in maven then the import will also be optional.
I will create another commit for this.

rakeshk15: to generate the OSGi metadata, the maven bundle plugin should be present in each pom
1. api
2. impl
3. each of json processor pom

not sure if Christian has some other way e.g. bnd maven plugin

cschneider: The settings in the parent are applied to all sub modules too. 
I checked the packages created and the optional imports above do not seem to be needed. So I removed this config.
As there might be the need to tweak individual modules later I added a config to support an optional bnd.bnd file per module. For now it is not needed but it makes sure the parent does not have to be changed later.

pom.xml
```diff
@@ -437,13 +435,7 @@
...
                 <configuration>
                     <instructions>
-                        <Import-Package><![CDATA[
-                            android.util;resolution:=optional,
-                            org.bouncycastle.jce;resolution:=optional,
-                            org.bouncycastle.jce.spec;resolution:=optional,
-                            *
-                        ]]>
-                        </Import-Package>
+                        <_include>-bnd.bnd</_include>
```
lhazlewood: Since I've never used this, I just want to confirm: 

Is this really `<_include>` (with an underscore) and not `<include>` ?  As well as `-bnd.bnd` and not `bnd.bnd` ?

cschneider: Yes. Not sure why they use the "_". The - in front of bnd.bnd means that the file is optional. So we can start without bnd.bnd files in the subprojects and add them only when something needs to be customized.

lhazlewood: Thanks!

lhazlewood:
Thanks so much for the PR!

Issue #171, opened 2016-09-16
Document whether DefaultJwtParser is thread-safe and/or resuable


dnault:
Looks like that's the intent; can you make a definitive statement?

Thanks!

lhazlewood:
Sure!  We hope to add proper docs soon (not just the README), so thanks for the issue!

agemooij:
Just to make sure this is in fact the case, could you please confirm that `DefaultJwtParser` is in fact thread-safe and that this is guaranteed to be the case in the future?

ricardoccpaiva:
any news on this ?

rubeninoto:
any news?

lhazlewood:
For those following along on this issue, JJWT the upcoming JJWT 0.11.0 release introduces a `JwtParserBuilder` that is returned from `Jwts.parserBuilder()` (contrast that with today's `Jwts.parser()`.  Instances of parsers returned from `Jwts.parserBuilder().....build()` are immutable.  Further, `Jwts.parser()` will be removed when 1.0.0 is released eventually.

cc @bdemers - can you please use this issue to ensure documentation for the 0.11.0 method JavaDoc, release notes and standard documentation makes this explicitly clear?

lhazlewood:
Closing per edits in #559.  Will be released with 0.11.0

PR #508, opened 2019-10-01
Docs: Add note about JwtParserBuilder creating an immutable JwtParser

bdemers:
Fixes: #171

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
+++ b/api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
@@ -298,8 +298,8 @@ public interface JwtParserBuilder {
     JwtParserBuilder deserializeJsonWith(Deserializer<Map<String,?>> deserializer);
 
     /**
-     * Returns a {@link JwtParser} created from the configuration from this JwtParserBuilder.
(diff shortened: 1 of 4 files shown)


api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
```diff
@@ -298,8 +298,8 @@
     JwtParserBuilder deserializeJsonWith(Deserializer<Map<String,?>> deserializer);
 
     /**
-     * Returns a {@link JwtParser} created from the configuration from this JwtParserBuilder.
-     * @return a JwtParser created from the configuration from this JwtParserBuilder.
+     * Returns a immutable/thread-safe {@link JwtParser} created from the configuration from this JwtParserBuilder.
+     * @return a immutable/thread-safe JwtParser created from the configuration from this JwtParserBuilder.
```
lhazlewood: Technically `an immutable/thread-safe`..., but I'll still approve 😉

bdemers: nice catch, i'll fix that before merging

PR #513, opened 2019-10-05
Cache Maven dependencies

cotsog:


lhazlewood:
@cotsog thanks for the PR, but can you please explain what this does or why it is beneficial?  To the best of my knowledge, Travis will fire up a new vm from scratch for each build in its build matrix, which means build N.1 won't use the same VM as build N.2.  If that's true, artifact caching doesn't make sense since after a build terminates, all local VM state is lost.

What am I missing? 😄

lhazlewood:
So I did some searching and found this:

https://docs.travis-ci.com/user/caching/#caching-directories-bundler-dependencies
and
https://docs.travis-ci.com/user/caching/#arbitrary-directories

If so, maybe they're using AWS EFS mount volumes - if so, that's awesome!  Anyway, it's worth a try, so thank you!

cotsog:
@lhazlewood Sorry for the lack of details on my PR. This is meant to make your builds more resilient to network connectivity issues.

We don't use volume mounts between builds. We rather compress your dependencies and upload them to a storage provider "close" (Google Cloud Storage) to the infrastructure where our VMs are running (Google Compute Engine). We then download the archive and extract its context at the beginning of the build.

Hope this helps!

lhazlewood:
@cotsog Awesome!  This is super cool - thanks so much for letting us know how it works, and thanks again for the PR!

PR #514, opened 2019-10-08
Remove extensions true, move jar plugin near bundle-plugin to show th…

cschneider:
…eir configs are related


lhazlewood:
Sorry, this build is consistently failing for JDK 7.  I think the version of the plugin you're using isn't compatible with that JDK.

lhazlewood:
Thanks!

PR #529, opened 2019-10-27
relocated Services.java to impl module (along with tests).

### LINKED ISSUE #524: Move Services class
from `io.jsonwebtoken.lang.Services` to the `impl` module in the package `io.jsonwebtoken.impl.lang.Services`.

dogeared:
resolves #524


lhazlewood:
LGTM!

lhazlewood:
Ah, JDK8 is having coverage failure tests.  Seems like something in the tests needs to be addressed.

dogeared:
Tests should be address @lhazlewood - we'll see after this build...

PR #540, opened 2019-12-16
Add DefaultJwtParser functionality to parse JWSs with empty body.

siebenfarben:
**Context**
We are currently implementing [RFC 8555](https://tools.ietf.org/html/rfc8555) which requires requests authenticated through JWS object payloads.
It suggests using signed [POST-AS-GET requests with empty JWS bodies](https://tools.ietf.org/html/rfc8555#page-13) for authenticating GET requests. When using jjwt library for the JWS handling we found that empty bodies are disallowed quite explicitly in `DefaultJwtParser` when parsing JWSs.

**Solution**
We added flag `allowEmptyBody`, alongside JwtParserBuilder fields and methods, to the DefaultJwtParser. This way the empty body check can still be used if required and allowing them is an intentional call.

To successfully parse empty-body JWS objects the parsed payload is accepted in a null-safe way (DefaultJWTParser lines 300 & 324). When proactively checking for and mapping json in line 335 we added a simple empty check for the payload.

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
+++ b/api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
@@ -39,7 +39,6 @@ public interface JwtParserBuilder {
      * Ensures that the specified {@code jti} exists in the parsed JWT.  If missing or if the parsed
      * value does not equal the specified value, an exception will be thrown indicating that the
      * JWT is invalid and may not be used.
-     * Sets payload required, {@link JwtParserBuilder#requirePayload(boolean)}.
      *
      * @param id
      * @return the parser builder for method chaining.
@@ -52,7 +51,6 @@ public interface JwtParserBuilder {
      * Ensures that the specified {@code sub} exists in the parsed JWT.  If missing or if the parsed
      * value does not equal the specified value, an exception will be thrown indicating that the
      * JWT is invalid and may not be used.
-     * Sets payload required, {@link JwtParserBuilder#requirePayload(boolean)}.
      *
      * @param subject
      * @return the parser builder for method chaining.
@@ -65,7 +63,6 @@ public interface JwtParserBuilder {
      * Ensures that the specified {@code aud} exists in the parsed JWT.  If missing or if the parsed
      * value does not equal the specified value, an exception will be thrown indicating that the
      * JWT is invalid and may not be used.
-     * Sets payload required, {@link JwtParserBuilder#requirePayload(boolean)}.
      *
      * @param audience
      * @return the parser builder for method chaining.
@@ -78,7 +75,6 @@ public interface JwtParserBuilder {
      * Ensures that the specified {@code iss} exists in the parsed JWT.  If missing or if the parsed
      * value does not equal the specified value, an exception will be thrown indicating that the
      * JWT is invalid and may not be used.
-     * Sets payload required, {@link JwtParserBuilder#requirePayload(boolean)}.
      *
      * @param issuer
      * @return the parser builder for method chaining.
@@ -91,7 +87,6 @@ public interface JwtParserBuilder {
      * Ensures that the specified {@code iat} exists in the parsed JWT.  If missing or if the parsed
      * value does not equal the specified value, an exception will be thrown indicating that the
      * JWT is invalid and may not be used.
-     * Sets payload required, {@link JwtParserBuilder#requirePayload(boolean)}.
      *
      * @param issuedAt
      * @return the parser builder for method chaining.
@@ -104,7 +99,6 @@ public interface JwtParserBuilder {
      * Ensures that the specified {@code exp} exists in the parsed JWT.  If missing or if the parsed
      * value does not equal the specified value, an exception will be thrown indicating that the
      * JWT is invalid and may not be used.
-     * Sets payload required, {@link JwtParserBuilder#requirePayload(boolean)}.
      *
      * @param expiration
      * @return the parser builder for method chaining.
@@ -117,7 +111,6 @@ public interface JwtParserBuilder {
      * Ensures that the specified {@code nbf} exists in the parsed JWT.  If missing or if the parsed
      * value does not equal the specified value, an exception will be thrown indicating that the
      * JWT is invalid and may not be used.
-     * Sets payload required, {@link JwtParserBuilder#requirePayload(boolean)}.
      *
      * @param notBefore
      * @return the parser builder for method chaining
@@ -130,7 +123,6 @@ public interface JwtParserBuilder {
      * Ensures that the specified {@code claimName} exists in the parsed JWT.  If missing or if the parsed
      * value does not equal the specified value, an exception will be thrown indicating that the
      * JWT is invalid and may not be used.
-     * Sets payload required, {@link JwtParserBuilder#requirePayload(boolean)}.
      *
      * @param claimName
      * @param value
@@ -139,16 +131,6 @@ public interface JwtParserBuilder {
      * @see IncorrectClaimException
      */
     JwtParserBuilder require(String claimName, Object value);
-
-    /**
-     * Require the payload of the JWT not to be empty. The default value is false. If a message contains an empty payload
-     * string but it is required per this flag, parsing the JWS will lead to a {@link MalformedJwtException}.
-     *
-     * @param payloadRequired
-     * @return the parser builder for method chaining.
-     * @see MalformedJwtException
-     */
-    JwtParserBuilder requirePayload(boolean payloadRequired);
 
     /**
      * Sets the {@link Clock} that determines the timestamp to use when validating the parsed JWT.
--- a/api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
+++ b/api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
@@ -133,6 +133,18 @@ public interface JwtParserBuilder {
     JwtParserBuilder require(String claimName, Object value);
 
     /**
+     * Allow that the body / payload part of the JWT may be empty. Some API specifications require JWT authenticated
+     * POST-AS-GET requests with an empty octet string as body part.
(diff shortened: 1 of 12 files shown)


lhazlewood:
cc @bdemers and @dogeared in case you guys want to chime in

api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
```diff
@@ -132,6 +132,18 @@
...
+     * Allow that the body / payload part of the JWT may be empty. Some API specifications require JWT authenticated
+     * POST-AS-GET requests with an empty octet string as body part.
+     * The default value is false. If a message contains an empty body without allowing it, parsing the JWS will lead
+     * to a {@link MalformedJwtException}.
+     *
+     * @param allowEmptyBody
+     * @return the parser builder for method chaining.
+     * @see MalformedJwtException
+     */
+    JwtParserBuilder allowEmptyBody(boolean allowEmptyBody);
```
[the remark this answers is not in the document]
siebenfarben: Hi @lhazlewood 
thanks for your comment and detailed explanation. Our understanding was as well that `MUST be present` includes an empty strings as a possibility.

I like your suggestion regarding `requirePayload()`. We will implement it and ping you here, when finished.

Best,
Philipp

siebenfarben:
Hi @lhazlewood 
please have a look at the updated PR. Thanks again for your comments!

api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java
```diff
@@ -132,6 +140,16 @@
...
 
+    /**
+     * Require the payload of the JWT not to be empty. The default value is false. If a message contains an empty payload
+     * string but it is required per this flag, parsing the JWS will lead to a {@link MalformedJwtException}.
+     *
+     * @param payloadRequired
+     * @return the parser builder for method chaining.
+     * @see MalformedJwtException
+     */
+    JwtParserBuilder requirePayload(boolean payloadRequired);
```
bdemers: We could probably remove the arg and make this just `requirePayload()`.  Assuming the default is always "false".  
@lhazlewood ?

bdemers:
Great stuff @siebenfarben !

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -293,7 +303,7 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
             base64UrlEncodedDigest = sb.toString();
         }
 
-        if (base64UrlEncodedPayload == null) {
+        if (base64UrlEncodedPayload == null && payloadRequired) {
```
lhazlewood: I was thinking about triggering exceptions due to `payloadRequired()` late last night and how an exception might manifest for the user.

Let's say someone configures `requireSubject("foo")` and then there isn't a payload.  The exception they'll see is a generic `MalformedJwtException` with a message `"JWT string '<whatever>' is missing a body/payload"`, and doesn't explain at all _why_ the JWT is required to have a payload.

To me, this seems to be undesirable.  If you think about it, in practice, almost no one really cares that there is a payload or not.  What they *really* care about is if something *in* the payload exists and/or matches an expectation.  And we already have support for enforcing and communicating those things via the other `require`* methods in the class.

(Also FWIW, I don't think `MalformedJwtException` is correct here - the JWT _isn't_ malformed - it's just that expectation/validation requirements have failed, so I think we'd have to throw another exception - probably `UnsupportedJwtException`, and that's if we throw at all...)

Which brings me to my next point:  do we need this check at all?

What if we didn't have a `payloadRequired` check at all and just let the empty string default as done on line 330 in this PR below?  The other `require*` methods would still fail with a meaningful/descriptive exception as is more desirable.

Would there be any problems?  Would any of the existing many tests in the project fail?  That would tell us a lot.

siebenfarben: Good point. 'Malformed' is certainly not the correct wording in that case. It's more an unmet requirement the callers asked for themselves.
All I can say from during the implementation - no tests fail if this check just goes away. The only thing that needs to be added would be the `isEmpty()` check for the payload below.

lhazlewood: Ok - in this case, I don't think we need the concept of `payloadRequired` at all and just keep the `isEmpty()` check.

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -317,6 +327,9 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
         }
 
         // =============== Body =================
+        if (base64UrlEncodedPayload == null) {
```
lhazlewood: if base64UrlEncodedPayload is `null`, we don't need to execute lines 333-337 below (no need to incur overhead of a Base64 Decoder and a CompressionCodec if there is nothing to decode or decompress).

Based on that, maybe the logic should look something like this:

```java
String payload = "";
if (base64UrlEncodedPayload != null) {
    byte[] bytes = base64UrlDecoder.decode(base64UrlEncodedPayload);
    if (compressionCodec != null) {
        bytes = compressionCodec.decompress(bytes);
    }
    payload = new String(bytes, Strings.UTF_8);
}
```

siebenfarben: Hi!
alright, I removed the `payloadRequired` methods and logic.

One last thing: There were each three tests in `JwtsTest` and `DeprecatedJwtsTest` regarding missing empty payloads:
* `testParseWithHeaderOnly` is now removed because given header `foo` is no valid base64 encoded header.
* `testParseWithHeaderAndSignatureOnly` is now removed because given header `foo` is no valid base64 encoded header and `bar` is no valid signature.
* `testParseWithTwoPeriodsOnly` is **ignored** for now, because while reading through the RFC I was not sure whether `..` is a valid JWT or not. What should we do about this?

Otherwise empty payloads are now supported. Thanks for your support. :-)

siebenfarben: (Unresolved conversation because of open question)

impl/src/test/groovy/io/jsonwebtoken/JwtsTest.groovy
```diff
@@ -171,33 +173,13 @@ class JwtsTest {
         }
     }
 
-    @Test
-    void testParseWithHeaderOnly() {
```
lhazlewood: We need to keep this test, since this amounts to an `Unsecured JWT` and they are supported.  If anything, instead of using `foo..` as the test string, we should use:

```groovy
Encoders.BASE64URL.encode('{"alg":"none"}') + '..'
```

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -294,7 +294,8 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
         }
 
         if (base64UrlEncodedPayload == null) {
-            throw new MalformedJwtException("JWT string '" + jwt + "' is missing a body/payload.");
+            // Payload string is empty.
+            base64UrlEncodedPayload = "";
```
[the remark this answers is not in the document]
siebenfarben: Alright. Sorry, I forgot to mention why I implemented it that way. Later the `jwtWithoutSignature` is built. The simple string concatenation leads to a `null` - which lets to errors later.
Please have a look at the null check I added for `jwtWithoutSignature` assignment.

impl/src/test/groovy/io/jsonwebtoken/JwtParserTest.groovy
```diff
@@ -168,6 +168,25 @@ class JwtParserTest {
         assertEquals jwt.body, payload
     }
 
+    @Test
+    void testParseEmptyBody() {
```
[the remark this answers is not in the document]
siebenfarben: Please have a look at the updated `JwtsBuilder` and the `testCompactWithoutPayloadOrClaims` test (I deleted one duplicate test).

siebenfarben:
Hi @lhazlewood, please have a look at the updated request. :) Small heads-up: I will from now on be on vacation without my laptop. Happy holidays! Best, @siebenfarben

impl/src/test/groovy/io/jsonwebtoken/DeprecatedJwtsTest.groovy
```diff
@@ -25,6 +25,7 @@ import io.jsonwebtoken.impl.lang.Services
 import io.jsonwebtoken.lang.Strings
 import io.jsonwebtoken.security.Keys
 import io.jsonwebtoken.security.WeakKeyException
+import org.junit.Ignore
```
lhazlewood: Please remove this now-unused import 😉

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -257,6 +257,11 @@ public Jwt parse(String jwt) throws ExpiredJwtException, MalformedJwtException,
 
         Assert.hasText(jwt, "JWT String argument cannot be null or empty.");
 
+        if ("..".equals(jwt)) {
```
lhazlewood: Why was this added?  Seems to very special case that is handled by standard processing later.

siebenfarben: My understanding in the discussion above was that `..` is not a valid JWS, because it needs to have at least a header.
Without this check, `..` is accepted.

siebenfarben:
I'm back - happy new year!
Are the changes now in state that can be merged? :)

lhazlewood:
Hi @siebenfarben - apologies for the long delay! This was lost in the cracks 😞 .  Merging now!

PR #545, opened 2020-01-16
Follow RFC of "alg" claim being case-sensitive

andrewwiik:
According to [https://tools.ietf.org/html/rfc7515#section-4.1.1](https://tools.ietf.org/html/rfc7515#section-4.1.1) the "alg" claim is case sensitive and should be treated as such.

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
+++ b/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
@@ -40,7 +40,7 @@ public enum SignatureAlgorithm {
     /**
      * JWA name for {@code No digital signature or MAC performed}
      */
-    NONE("none", "No digital signature or MAC performed", "None", null, false, 0, 0),
+    none("none", "No digital signature or MAC performed", "None", null, false, 0, 0),
 
     /**
      * JWA algorithm name for {@code HMAC using SHA-256}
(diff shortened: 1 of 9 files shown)


lhazlewood:
Hi there! Thanks for the PR!  However, I think the current codebase is fine.

1.  We can't change the case of the `SignatureAlgorithm.NONE` enum constant:
    1. This is a backwards incompatible change that would break existing user code if they upgrade JJWT.
    1. Per semantic versioning rules, we can only make this kind of change on a major version upgrade - e.g. 1.0.0. However that's not really relevant because: 
    1.  The String `value` assigned to the enum is already the lowercase `none` as the spec uses.
    1.  It is idiomatic Java to use uppercase to reflect contstants, which is why `SignatureAlgorithm.NONE` is uppercase, but it's string `value` is lowercase per the spec.
1.  Except where security may be compromised, JJWT _usually_ tries to follow the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle) where it is flexible on what it accepts (e.g. SignatureAlgorithm line 632's `.equalsIgnoreCase`), but it is strict on what it produces (JJWT will always write a lowercase claim for `none`, not `None` or `NONE`).

We welcome PRs!  However, [per our documentation, please discuss proposed changes first](https://github.com/jwtk/jjwt#contributing-pull-requests) to ensure that your don't spend time creating a PR that might not be accepted.  Thanks!

PR #546, opened 2020-01-17
Add SpotBugs / Find Security Bugs

bdemers:
Sets the default encoding in `GsonDeserializer` to `UTF_8`
Previously this was unset, and used the system default

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/lang/Classes.java
+++ b/api/src/main/java/io/jsonwebtoken/lang/Classes.java
@@ -18,6 +18,8 @@ package io.jsonwebtoken.lang;
 import java.io.InputStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 
 /**
  * @since 0.1
@@ -193,10 +195,18 @@ public final class Classes {
     public static <T> T invokeStatic(String fqcn, String methodName, Class[] argTypes, Object... args) {
         try {
             Class clazz = Classes.forName(fqcn);
-            Method method = clazz.getDeclaredMethod(methodName, argTypes);
-            method.setAccessible(true);
+            final Method method = clazz.getDeclaredMethod(methodName, argTypes);
+
+            AccessController.doPrivileged(new PrivilegedAction<Object>() {
+                @Override
+                public Object run() {
+                    method.setAccessible(true);
+                    return null;
+                }
+            });
+
             return(T)method.invoke(null, args);
-        } catch (Exception e) {
+        } catch (Exception e) { // TODO catch specific exceptions (avoid catching RuntimeException), breaking change fix in 1.0
             String msg = "Unable to invoke class method " + fqcn + "#" + methodName + ".  Ensure the necessary " +
                 "implementation is in the runtime classpath.";
             throw new IllegalStateException(msg, e);
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
@@ -99,7 +99,9 @@ public class DefaultJwtParser implements JwtParser {
                      CompressionCodecResolver compressionCodecResolver) {
         this.signingKeyResolver = signingKeyResolver;
         this.key = key;
-        this.keyBytes = keyBytes;
+        if (keyBytes != null) {
+            this.keyBytes = keyBytes.clone();
+        }
         this.clock = clock;
         this.allowedClockSkewMillis = allowedClockSkewMillis;
         this.expectedClaims = expectedClaims;
@@ -188,7 +190,7 @@ public class DefaultJwtParser implements JwtParser {
     @Override
     public JwtParser setSigningKey(byte[] key) {
         Assert.notEmpty(key, "signing key cannot be null or empty.");
-        this.keyBytes = key;
+        this.keyBytes = key.clone();
         return this;
     }
 
@@ -485,9 +487,10 @@ public class DefaultJwtParser implements JwtParser {
 
     private void validateExpectedClaims(Header header, Claims claims) {
 
-        for (String expectedClaimName : expectedClaims.keySet()) {
+        for (Map.Entry<String, ?> expectedClaimEntry : expectedClaims.entrySet()) {
 
-            Object expectedClaimValue = normalize(expectedClaims.get(expectedClaimName));
+            String expectedClaimName = expectedClaimEntry.getKey();
+            Object expectedClaimValue = normalize(expectedClaimEntry.getValue());
             Object actualClaimValue = normalize(claims.get(expectedClaimName));
 
             if (expectedClaimValue instanceof Date) {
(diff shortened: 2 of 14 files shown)


bdemers:
Wow @lhazlewood, I had convinced myself 0.11 was already out. 🤦‍♂ 

I changed the note to say that Gson support was added.

As for the test, I added SpotBugs and Find Security Bugs (which is how I found this initially)
Added a few fixes for those issues as well, a few of them are less critical but were simple (and easier to fix than exclude)

Thoughts? (spotbugs changes are in a08c1fba109aa3335f38fd782d2f9c78ff5d7e75)

lhazlewood:
I meant an actual unit test that inspects the `String` instance and ensures the encoding is `UTF-8`.

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -99,7 +99,9 @@ public DefaultJwtParser() { }
                      CompressionCodecResolver compressionCodecResolver) {
         this.signingKeyResolver = signingKeyResolver;
         this.key = key;
-        this.keyBytes = keyBytes;
+        if (keyBytes != null) {
+            this.keyBytes = keyBytes.clone();
```
lhazlewood: Why was this added?  Could be a minor performance hit.

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -188,7 +190,7 @@ public JwtParser setAllowedClockSkewSeconds(long seconds) {
     @Override
     public JwtParser setSigningKey(byte[] key) {
         Assert.notEmpty(key, "signing key cannot be null or empty.");
-        this.keyBytes = key;
+        this.keyBytes = key.clone();
```
lhazlewood: Same.

api/src/main/java/io/jsonwebtoken/lang/Classes.java
```diff
@@ -193,10 +196,18 @@ public static boolean isAvailable(String fullyQualifiedClassName) {
...
+                @Override
+                public Object run() {
+                    method.setAccessible(true);
+                    return null;
+                }
+            });
+
             return(T)method.invoke(null, args);
-        } catch (Exception e) {
+        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
```
lhazlewood: This is not backwards-compatible functionality, we can't really do this until 1.0, right?  If any other RuntimeException is thrown, it won't be caught (like it is today).

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParserBuilder.java
```diff
@@ -138,7 +138,7 @@ public JwtParserBuilder setAllowedClockSkewSeconds(long seconds) {
     @Override
     public JwtParserBuilder setSigningKey(byte[] key) {
         Assert.notEmpty(key, "signing key cannot be null or empty.");
-        this.keyBytes = key;
+        this.keyBytes = key.clone();
```
lhazlewood: Is this necessary?

impl/src/main/java/io/jsonwebtoken/impl/crypto/EllipticCurveProvider.java
```diff
@@ -140,7 +143,7 @@ public static KeyPair generateKeyPair(String jcaAlgorithmName, String jcaProvide
             ECGenParameterSpec spec = new ECGenParameterSpec(paramSpecCurveName);
             g.initialize(spec, random);
             return g.generateKeyPair();
-        } catch (Exception e) {
+        } catch (InvalidAlgorithmParameterException | NoSuchAlgorithmException | NoSuchProviderException e) {
```
lhazlewood: Same comment about RuntimeExceptions that would be missed but are caught today.

impl/src/main/java/io/jsonwebtoken/impl/crypto/EllipticCurveSignatureValidator.java
```diff
@@ -50,7 +50,7 @@ public boolean isValid(byte[] data, byte[] signature) {
              * **/
             byte[] derSignature = expectedSize != signature.length && signature[0] == 0x30 ? signature : EllipticCurveProvider.transcodeSignatureToDER(signature);
             return doVerify(sig, publicKey, data, derSignature);
-        } catch (Exception e) {
+        } catch (InvalidKeyException | java.security.SignatureException e) {
```
lhazlewood: idem

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
bdemers: to prevent `keyBytes` from mutating at some point later on. The parser is meant to be reused, so perf _shouldn't_ be an issue, but... YMMV

api/src/main/java/io/jsonwebtoken/lang/Classes.java
bdemers: Technically, but I think this one might be a far edge case, a RuntimeException would have to be thrown, and the caller would need to be catching an `IllegalStateException`.

bdemers:
Probably just change the name, but let's wrap up the other open questions/issues first and then figure out the best option

api/src/main/java/io/jsonwebtoken/lang/Classes.java
lhazlewood: True - but it is possible given the number of projects that use jjwt.  I'd feel safer if we don't change this until 1.0.

lhazlewood: Could you add a //TODO on the three places and add a follow-up GitHub issue for us to remove it for 1.0 (and add it to the 1.0 release tag?)

bdemers:
@lhazlewood have a suggestion on a good way to do this?
> I meant an actual unit test that inspects the String instance and ensures the encoding is UTF-8.

The only option I could think of was to create a byte array with characters known to render differently, but IMHO, it's easier to scan the source tree for these types of issues 🤔

lhazlewood:
@bdemers just create a mock gson instance and mock `gson.fromJson`, and assert that the argument is UTF-8 compatible?

I found this: https://www.turro.org/publications/?item=114&page=0

But if you don't want to mess with that, no objection from me.  I'm just paranoid 😅

bdemers:
I was messing with creating keys today with BC (via JCA), and it's really really easy to get a `RuntimeException` to be thrown.  I think if we just tack on a `| RuntimeException` we should be good for backward compatibility.

lhazlewood:
> I think if we just tack on a `| RuntimeException` we should be good for backward compatibility.

If that's the case, why change the code at all?  Just curious.

bdemers:
Valid point, I was just thinking it would be easier to change in the future, but it's probably just noise

bdemers:
@lhazlewood back to the topic of the test, we actually do have tests that fail when the system's default encoding is NOT `UTF-8` we just don't see this because that is usually what Linux and MacOS is set to.

Adding a test for this is challenging because it would require forking a JVM to change the `file.encoding` property.

I tried this from my IDE, but you _should_ be able to reproduce it by running the tests against `master` with following arg`-Dfile.encoding=ISO-8859-1`

bdemers:
I updated the PR name to better reflect the changes here, and I've moved the Gson change log note to: #547

lhazlewood:
@bdemers what do you want to do with this? Thoughts?

bdemers:
I'll take another pass at this in a different PR, based off the latest code

PR #554, opened 2020-02-04
SignatureAlgorithm PS256, PS384, and PS512 on Java 11+ without BouncyCastle

### LINKED ISSUE #542: Support for PS256 in Java 11
According to the documentation starting from Java 11 it should be possible to use signatures with PS256 algorithm, but when I try this it gives an error unless BouncyCastle is used. But I though that starting from Java 11 BouncyCastle would just not be necessary anymore.
Any tips?
`.signWith(signKey, SignatureAlgorithm.PS256)`
gives
`io.jsonwebtoken.security.SignatureException: Unavailable RSA Signature algorithm 'SHA256withRSAandMGF1'. This is not a standard JDK algorithm. Try including BouncyCastle in the runtime classpath.
`
Looks like by using PS256 it tries Signature algorithm SHA256withRSAandMGF1 which seems BouncyCastle specific ?

lhazlewood:
Ensured SignatureAlgorithm `PS256`, `PS384`, and `PS512` works properly on JDK 11 and later without requiring BouncyCastle.  Resolves #542


lhazlewood:
cc @fschutte

PR #566, opened 2020-02-24
Allow JacksonDeserializer to work with Java 9+ Map.of

bdemers:
and other maps that do NOT allow null keys

### DIFF
--- a/extensions/jackson/src/main/java/io/jsonwebtoken/jackson/io/JacksonDeserializer.java
+++ b/extensions/jackson/src/main/java/io/jsonwebtoken/jackson/io/JacksonDeserializer.java
@@ -120,8 +120,9 @@ public class JacksonDeserializer<T> implements Deserializer<T> {
         @Override
         public Object deserialize(JsonParser parser, DeserializationContext context) throws IOException {
             // check if the current claim key is mapped, if so traverse it's value
-            if (claimTypeMap != null && claimTypeMap.containsKey(parser.currentName())) {
-                Class type = claimTypeMap.get(parser.currentName());
+            String name = parser.currentName();
+            if (claimTypeMap != null && name != null && claimTypeMap.containsKey(name)) {
+                Class type = claimTypeMap.get(name);
                 return parser.readValueAsTree().traverse(parser.getCodec()).readValueAs(type);
             }
             // otherwise default to super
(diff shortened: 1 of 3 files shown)


extensions/jackson/src/test/groovy/io/jsonwebtoken/jackson/io/JacksonDeserializerTest.groovy
```diff
@@ -105,6 +105,41 @@ class JacksonDeserializerTest {
...
+                    throw new NullPointerException("key is null, expected for this test")
+                }
+                return super.containsKey(key)
+            }
+        }
+
+        // TODO: the following does NOT work with Java 1.7
+        // when we stop supporting that version we can use a partial mock instead
+        // the `typeMap.put("custom", CustomBean)` line below results in an NPE, (only on 1.7)
+
```
bdemers: @lhazlewood any ideas on how to clean up this test?
The partial mock doesn't work on 1.7, and I'm not sure I want to dig into that, but I feel like the above solution could be better too...

lhazlewood: I don't know about the `partialMockBuilder` myself - never used it.  That said, I like creating the anonymous subclass - that's even better than a mock IMO.

extensions/jackson/src/main/java/io/jsonwebtoken/jackson/io/JacksonDeserializer.java
```diff
@@ -120,7 +120,8 @@ private MappedTypeDeserializer(Map<String, Class> claimTypeMap) {
         @Override
         public Object deserialize(JsonParser parser, DeserializationContext context) throws IOException {
             // check if the current claim key is mapped, if so traverse it's value
-            if (claimTypeMap != null && claimTypeMap.containsKey(parser.currentName())) {
+
+            if (claimTypeMap != null && parser.currentName() != null && claimTypeMap.containsKey(parser.currentName())) {
```
lhazlewood: Should we memoize `parser.currentName()` ?  It's called in this code 3 different times.

lhazlewood:
Also, don't forget to target a 0.11.x branch

bdemers:
Agreed, lets get the 0.11.1 release out with this and the Gson issue out, and then cut a maintenance branch for 0.11.x so we can continue with 1.0

Issue #568, opened 2020-03-03
ServiceLocator can't find an implementation in common-pool workers


timsazon:
Hi, I've noticed that we can't read/create a token in a worker from `ForkJoinPool.commonPool()` in Spring's uber-jar application because ServiceLocator can't find an implementation for that (looks like there is a problem with context class loader related to the thread).

Example (it will throw `UnavailableImplementationException` for `Serializer.class`):
```
ForkJoinPool.commonPool()
        .execute(() -> Jwts.builder()
                .setClaims(Map.of(Claims.SUBJECT, "subject"))
                .compact());
```

Currently, I have a workaround to that by providing my own serializer (like this https://github.com/jwtk/jjwt#jackson-json-processor).

lhazlewood:
Interesting.  Do you have a simple test case that would allow us to re-create this?  Or even a minimal set of code to get us started?

Also, have you seen this?

https://stackoverflow.com/questions/49113207/completablefuture-forkjoinpool-set-class-loader

I wonder if that or something similar could help.

cc @bdemers

timsazon:
Thank you for the link!

Unfortunately, I can't get rid of `ForkJoinPool` in my case because of it in use by Caffeine's cache loader (I'll check the configuration). 

Btw, I made an example app: https://github.com/timsazon/jjwt-568

bdemers:
Hey @timsazon thanks for putting together a minimal example, that made things MUCH easier.

For a short term workaround you could do something like this:

```java
    private static Supplier<String> tokenSupplier() {
        return () -> {
            ClassLoader originalClassloader = Thread.currentThread().getContextClassLoader();
            try {
                Thread.currentThread().setContextClassLoader(Jwts.class.getClassLoader());
                return Jwts.builder()
                        .setClaims(Map.of(Claims.SUBJECT, "subject"))
                        .compact();
            } finally {
                Thread.currentThread().setContextClassLoader(originalClassloader);
            }
        };
    }
```
Basically just setting the contextClassLoader to something with JJWTs classes.

(or if you can configure the thread pool as mentioned in that stackoverflow post

@lhazlewood 
We _could_ set the classloader in Services: https://github.com/jwtk/jjwt/blob/master/impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java#L71
But, we don't actually know which classloader to use (in the case of a more complex classloader setup)

We _could_ also try to fall back, something like this in Services:

```java
[...]

PR #574, opened 2020-03-11
Services now checks the contextClassLoader, Services.class.classLoader, and the system classloader

bdemers:
Services are now loaded from the context class loader, Services.class.classLoader, and the system classloader, the first classloader with a service wins, and the others are ignored. This mimics how `Classes.forName()` works, and how JJWT attempted to auto-discover various implementations in previous versions.

Fixes: #568

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java
@@ -22,11 +22,34 @@ import java.util.Collections;
 import java.util.List;
 import java.util.ServiceLoader;
 
+import static io.jsonwebtoken.lang.Collections.arrayToList;
+
 /**
  * Helper class for loading services from the classpath, using a {@link ServiceLoader}. Decouples loading logic for
  * better separation of concerns and testability.
  */
 public final class Services {
+
+    private static final List<ClassLoaderAccessor> CLASS_LOADER_ACCESSORS = arrayToList(new ClassLoaderAccessor[] {
+            new ClassLoaderAccessor() {
+                @Override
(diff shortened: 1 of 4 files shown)


impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java
```diff
@@ -22,12 +22,35 @@
...
 
+import static io.jsonwebtoken.lang.Collections.arrayToList;
+
 /**
  * Helper class for loading services from the classpath, using a {@link ServiceLoader}. Decouples loading logic for
  * better separation of concerns and testability.
  */
 public final class Services {
 
+    private static final List<ClassLoaderAccessor> CLASS_LOADER_ACCESSORS = arrayToList(new ClassLoaderAccessor[] {
```
lhazlewood: I think this class would be much cleaner and reduce cyclomatic complexity / loops if we wrapped the list in our own delegating ClassLoader implementation.   Thoughts?

lhazlewood:
It probably goes without saying, but I'll ask anyway - were you able to verify this works w/ @timsazon's sample app?

lhazlewood:
(didn't see your comment in the Issue until just now :-P )

impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java
bdemers: I was going to go down that path, but I figured this option was less complex. Mostly because I'm unaware of any gotchas (ClassLoader is a 3000+ abstract class).  I'm guessing the SecurityManager could come into play 🤷‍♂

PR #580, opened 2020-03-19
Attach implementation bundles as fragments to the API bundle to make …

cziegeler:
…them visible to the API bundle. This fixes jwtk#578 (jwtk#578)


lhazlewood:
@cziegeler sorry I missed this. Is this the only thing necessary for OSGi to work again?

cziegeler:
@lhazlewood Yes, nothing else is needed

Issue #583, opened 2020-03-26
Reject seconds values that would cause overflow


jonalbertini:
If seconds used with long max value, there will be an overflow because you put the value (long) multiplied by 1000 in another long...
in my case i got a negative value as result and the Math.max then return 0....

as I was expecting to have the max ClockSkew, I got the Min....

https://github.com/jwtk/jjwt/blob/dacdb2ce25d1161c03904a874da88107b0459766/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParserBuilder.java#L134

lhazlewood:
I don't quite understand the problem.  You shouldn't have overflow with any relevant value, correct?  Clock skew should ideally be no more than 5 to 10 minutes at most (e.g. 360 seconds).  Multiplying 360 * 1000 shouldn't ever cause overflow.

I'm curious - how did this happen?  What was your input value?

jonalbertini:
Hello,
 the problem is not in what the clock skew should be. you can't predict the
"use cases". you deliver a function, doing something , with an Input and an
effect. In my case , the framework I use, in TEST cases use the
setAllowedClockSkew to allow expired token. It seems that's the only way to
allow it. so we put the max long Value in input of your function. but the
behaviour is not the one we want...
you expose a function with a long in input, and multiply this input by 1000
. this value is then stored in a long. even il we do abstraction of what
the function does, it's an issue. if you assume that there's a max value
allowed you have to specify it, document it, and code some verifications.

if you do so and really limit the clock skew to the "ideally" value. you
may expose a setting to avoid the clock skew control for tests.

my input value il the long max positive value.

best regards,
you already di an awesome job with this lib. we use it a lot in production.
thank you .

Le ven. 27 mars 2020 à 21:57, Les Hazlewood <notifications@github.com> a
écrit :

lhazlewood:
Test cases written without any insight into why an API exists, or to use the API in an unintended manner, will cause problems like this.  In this case, these test cases are effectively using a feature of the library to avoid the designed purpose of the method - to account for clock skew during parsing.  This method does not exist to allow parsing expired tokens, so the test is using it incorrectly.

If anything, a nicer solution is probably to reject any value greater than (`Long.MAX_VALUE / 1000`) to guarantee that overflow is not possible.

[...]

PR #584, opened 2020-03-27
Allowed clock skew millis

jonalbertini:
Duplicate method with millis to avoid overflow is max values are used


lhazlewood:
Thanks for the PR!  However, per the discussion in #583, we'd like to use a different design (input assertions) instead of adding another method to the public API (since that method will be removed quickly in favor of `java.time.Duration` arguments at a later date).

Issue #588, opened 2020-04-26
SignatureAlgorithm#assertValid should allow OIDs as well as JCA standard names


lhazlewood:
Per https://github.com/jwtk/jjwt/issues/381#issuecomment-619302910: 

In addition to the existing jcaName checks for JCA Standard Names, also allow OIDs as well due to PKCS12 keystore implementations that use them instead of standard names.

lhazlewood:
Resolved via #589.  Will be released in 0.11.2

PR #589, opened 2020-05-02
Accept oid in assertValid

fibsifan:
Hello,

This is a first draft for jwtk/jjwt#588

I've added the OIDs for Hmac-Keys and extended the check in the assertValid method. There are multiple ways of doing this, I decided to extend the SignatureAlgorithm's constructor with a varargs parameter to avoid having to modify all enum values.

For the test, an actual pkcs12 KeyStore instance is used to store and load the key. I suppose, one could mock the key. Would that be necessary?

The coverage checks failed to upload in my cloned repository, I suppose they will successfully do so, when the pr is built.

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
+++ b/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
@@ -123,18 +123,20 @@ public enum SignatureAlgorithm {
     private final int digestLength;
     private final int minKeyLength;
     /**
+     * Algorithm name as given by {@link Key#getAlgorithm()} if the key was loaded from a pkcs12 Keystore.
+     *
      * @deprecated This is just a workaround for https://bugs.openjdk.java.net/browse/JDK-8243551
      */
     @Deprecated
-    private final String oid;
+    private final String pkcs12Name;
 
     SignatureAlgorithm(String value, String description, String familyName, String jcaName, boolean jdkStandard,
                        int digestLength, int minKeyLength) {
-        this(value, description,familyName, jcaName, jdkStandard, digestLength, minKeyLength, null);
+        this(value, description,familyName, jcaName, jdkStandard, digestLength, minKeyLength, jcaName);
     }
 
     SignatureAlgorithm(String value, String description, String familyName, String jcaName, boolean jdkStandard,
-                       int digestLength, int minKeyLength, String oid) {
+                       int digestLength, int minKeyLength, String pkcs12Name) {
         this.value = value;
         this.description = description;
         this.familyName = familyName;
@@ -142,7 +144,7 @@ public enum SignatureAlgorithm {
         this.jdkStandard = jdkStandard;
         this.digestLength = digestLength;
         this.minKeyLength = minKeyLength;
-        this.oid = oid;
+        this.pkcs12Name = pkcs12Name;
     }
 
     /**
@@ -365,9 +367,9 @@ public enum SignatureAlgorithm {
             if (!HS256.jcaName.equalsIgnoreCase(alg) &&
                 !HS384.jcaName.equalsIgnoreCase(alg) &&
                 !HS512.jcaName.equalsIgnoreCase(alg) &&
-                !HS256.oid.equals(alg) &&
-                !HS384.oid.equals(alg) &&
-                !HS512.oid.equals(alg)) {
+                !HS256.pkcs12Name.equals(alg) &&
+                !HS384.pkcs12Name.equals(alg) &&
+                !HS512.pkcs12Name.equals(alg)) {
                 throw new InvalidKeyException("The " + keyType(signing) + " key's algorithm '" + alg +
                     "' does not equal a valid HmacSHA* algorithm name and cannot be used with " + name() + ".");
             }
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
```diff
@@ -122,16 +122,18 @@
     private final boolean jdkStandard;
     private final int digestLength;
     private final int minKeyLength;
+    private final List<String> alternativeNames;
```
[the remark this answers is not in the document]
lhazlewood: https://hg.openjdk.java.net/jdk/jdk/rev/62d97ec739db#l26.2

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
```diff
@@ -289,6 +291,20 @@ public int getMinKeyLength() {
...
+     * Returns alternative algorithm names to the {@link #getJcaName() jcaName} that may occur, e.g. when loading keys
+     * from a pkcs#12 keystore.
+     *
+     * @return alternative algorithm names to the {@link #getJcaName() jcaName} that may occur, e.g. when loading keys
+     * from a pkcs#12 keystore.
+     * @since 0.11.2
+     * @deprecated To be removed, when <a href="https://bugs.openjdk.java.net/browse/JDK-8243551">JDK-8243551</a> is fixed.
+     */
+    @Deprecated
+    public List<String> getAlternativeNames() {
```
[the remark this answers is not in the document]
fibsifan: I've removed the public method in 9589a57.

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
```diff
@@ -122,16 +122,27 @@
...
     private final int minKeyLength;
+    /**
+     * @deprecated This is just a workaround for https://bugs.openjdk.java.net/browse/JDK-8243551
+     */
+    @Deprecated
+    private final String oid;
 
     SignatureAlgorithm(String value, String description, String familyName, String jcaName, boolean jdkStandard,
                        int digestLength, int minKeyLength) {
+        this(value, description,familyName, jcaName, jdkStandard, digestLength, minKeyLength, null);
```
lhazlewood, line 133: Instead of passing `null`, is it possible to use the OIDs listed in the OpenJDK link I provided for the other algorithms?  Thoughts?

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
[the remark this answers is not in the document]
fibsifan: I have extended the test I used to demonstrate this error to check all signature-algorithms against the JCA and Bouncycastle-Keystores. Only HmacSha-Keys loaded from a JCA-pkcs12-Keystore have shown the mentioned behaviour, so I went with checking the OID only for the HmacSha SignatureAlgorithms for now.

The following table lists Jca Names for each (private) Key when loaded from each keystore:

SignatureAlgorithm / Keystore | jca        | pkcs12              | jks | jceks      | bks        | uber       | bcpkcs12
------------------------------|------------|---------------------|-----|------------|------------|------------|---------
HS256                         | HmacSHA256 | 1.2.840.113549.2.9  |     | HmacSHA256 | HmacSHA256 | HmacSHA256 |         
HS384                         | HmacSHA384 | 1.2.840.113549.2.10 |     | HmacSHA384 | HmacSHA384 | HmacSHA384 |         
HS512                         | HmacSHA512 | 1.2.840.113549.2.11 |     | HmacSHA512 | HmacSHA512 | HmacSHA512 |         
RS256                         | RSA        | RSA                 | RSA | RSA        | RSA        | RSA        | RSA     
RS384                         | RSA        | RSA                 | RSA | RSA        | RSA        | RSA        | RSA     
RS512                         | RSA        | RSA                 | RSA | RSA        | RSA        | RSA        | RSA     
ES256                         | EC         | EC                  | EC  | EC         | EC         | EC         | EC      
ES384                         | EC         | EC                  | EC  | EC         | EC         | EC         | EC      
ES512                         | EC         | EC                  | EC  | EC         | EC         | EC         | EC      
PS256                         | RSA        | RSA                 | RSA | RSA        | RSA        | RSA        | RSA     
PS384                         | RSA        | RSA                 | RSA | RSA        | RSA        | RSA        | RSA     
PS512                         | RSA        | RSA                 | RSA | RSA        | RSA        | RSA        | RSA

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
fibsifan: Sorry, I took a while: https://github.com/jwtk/jjwt/pull/589#discussion_r436289039

Seems, that it's not necessary to have oids for every Signature-Algorithm. Should we have them anyways? Or could we leave them out, since the field is going to be removed anyways?

lhazlewood: Ah, I see!  Instead of `oid` then, we should have a `pkcs12Name` field (instead of `oid`) that captures the names represented in the table above.  This guarantees that everything is non-null as well as ensures a parallel variable (`jcaName` and `pkcs12Name` make more sense now).

fibsifan: Yes, I agree. I renamed the field and made it a copy of jcaName for non-HmacSHA algorithms.

fibsifan:
Now the field pkcs12Name has always a value (except for `SignatureAlgorithm.NONE`), but the fallback namecheck is only performed for hmac-algirithms.

lhazlewood:
Resolves #588

PR #590, opened 2020-05-09
Add support for jdk14

fibsifan:
Hi,

I've noticed, that jjwt tests fail in JDK14 because of [GROOVY-9211]. I've taken the liberty to add openjdk14 as a build in travis configuration and bumped the groovy version to 2.5.11, where the issue is fixed.

[GROOVY-9211]: https://issues.apache.org/jira/browse/GROOVY-9211


lhazlewood:
Thank you for the PR!

PR #609, opened 2020-07-09
Removes Java 7 from CI test matrix

bdemers:
🤔 once Java 15 drops, we _should_ probably update the test to run against 8, 11, latest (15). So we don't run into issues where it becomes difficult to install the older unsupported JVMs


bdemers:
Updated Java 7 support in #610

Issue #578, opened 2020-03-18
Unable to find implementation classes in OSGi


cziegeler:
At least since the removal of the service loader for loading implementations, the library does not work in OSGi anymore. The reason is that the api bundle does not have access to classes from the impl bundle as class loading across bundles only works if those classes are exported (and imported).
I'm not sure how to solve this; clearly it would be wrong to export implementation classes and import them in the api.
Creating a single bundle with api and impl would work, with the expense of creating another artifact.

This is the exception I get with latest 0.11.1 :

io.jsonwebtoken.lang.UnknownClassException: Unable to load class named [io.jsonwebtoken.impl.DefaultJwtBuilder] from the thread context, current, or system/application ClassLoaders.  All heuristics have been exhausted.  Class could not be found.  Have you remembered to include the jjwt-impl.jar in your runtime classpath?
        at io.jsonwebtoken.lang.Classes.forName(Classes.java:92)
        at io.jsonwebtoken.lang.Classes.newInstance(Classes.java:136)
        at io.jsonwebtoken.Jwts.builder(Jwts.java:141)
        at de.osoco.cziegeler.test.jjwt.Test.start(Test.java:12)

lhazlewood:
@cziegeler thanks for the issue - I really appreciate it.

Some questions:

Did this problem happen to you when using 0.10.x or earlier?

Could it be that we're not exporting the OSGi metadata correctly?  This is the only thing we have for maintaining OSGi bundle data:

https://github.com/jwtk/jjwt/blob/c09deaa5f31fb7b3a8184daec4228269942c77a4/pom.xml#L485-L503

(This is in <build><plugins> to ensure it is enabled for all children modules automatically).

Thoughts?  cc @bdemers

bdemers:
Hi @cziegeler!

cziegeler:
@lhazlewood Yes, this happended with the 10.x version as well; but I only tested with the latest 10.x release.
@bdemers The OSGi metadata is correct - the api jar exports everything and the impl jar does not export anything - this is how it should be. But due to that the api does not find the impl.
I have some ideas on how to fix it, I'll try them out and then report back here

cziegeler:
Just some more information: the first official release supporting OSGi is 0.11.0 (which has the same problem).
The easiest solution is to make the implementation bundles fragments and therefore make them visible to the api bundle. In this case, it works exactly as it does outside of OSGi.
I've created a PR for this: #580

cziegeler:
[...]

PR #611, opened 2020-07-09
Removes eager initialization of the CompressionCodecResolver in the JwtParserBuilder

bdemers:
This addresses a potential service loader issue with OSGi runtimes.

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParserBuilder.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParserBuilder.java
@@ -55,7 +55,7 @@ public class DefaultJwtParserBuilder implements JwtParserBuilder {
 
     private SigningKeyResolver signingKeyResolver;
 
-    private CompressionCodecResolver compressionCodecResolver = new DefaultCompressionCodecResolver();
+    private CompressionCodecResolver compressionCodecResolver;
 
     private Decoder<String, byte[]> base64UrlDecoder = Decoders.BASE64URL;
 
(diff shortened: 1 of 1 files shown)


bdemers:
The claim in #578 is that OSGi has issues loading Services (I have an open question in that issue to see if there is something _better_ we can do there).

Specifically the comment from @lexsoto: https://github.com/jwtk/jjwt/issues/578#issuecomment-655728917

This change lazy loads the `CompressionCodecResolver` because the constructor of the [`DefaultCompressionCodecResolver`](https://github.com/jwtk/jjwt/blob/dacdb2ce25d1161c03904a874da88107b0459766/impl/src/main/java/io/jsonwebtoken/impl/compression/DefaultCompressionCodecResolver.java#L58-L68) eagerly loads services.

https://github.com/jwtk/jjwt/blob/dacdb2ce25d1161c03904a874da88107b0459766/impl/src/main/java/io/jsonwebtoken/impl/compression/DefaultCompressionCodecResolver.java#L58-L68

lhazlewood:
In that same thread @lexsoto [referenced](https://github.com/jwtk/jjwt/issues/578#issuecomment-655064450) this:

https://blog.osgi.org/2013/02/javautilserviceloader-in-osgi.html

That article indicates how, via metadata only (META-INF/services and META-INF/MANIFEST.MF), OSGi environments can enable `ServiceLoader` functionality.

Since we control both the 'provider' .jars (the Jackson, Gson, etc implementations) and the 'consumer' jar (the `impl` module that creates `DefaultCompressionCodecResolver`), can't we just solve this with metadata and without code changes?

@lexsoto and @cziegeler - would either of you be able to help us set up Karaf in an `OsgiIT` integration test and test this somehow so we can ensure this stuff works on every build?

Unless we have an automated repeatable solution, this problem is likely to surface repeatedly.

lhazlewood:
Continuing this discussion in the Issue thread as it's not really related to this PR:

https://github.com/jwtk/jjwt/issues/578#issuecomment-657086933

PR #620, opened 2020-09-07
Remove redundant Proguard statements

shinwan2:
Proguard `keepnames` is redundant with `keep`. 

Based on [Proguard manual](https://www.guardsquare.com/en/products/proguard/manual/usage), `interface` modifier is redundant because `class` modifier already covers classes and interfaces.
> The class keyword refers to any interface or class. Theinterface keyword restricts matches to interface classes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,11 +336,8 @@ You can use the following [Android Proguard](https://developer.android.com/studi
 -keepattributes InnerClasses
 
 -keep class io.jsonwebtoken.** { *; }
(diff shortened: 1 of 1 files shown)


lhazlewood:
Is this true for older version of Android?  We need to ensure it works for them as well.

lhazlewood:
No answer from OP - closing.  Please open a new issue (or update this PR and comment on it) if you'd like it to go in.  Please ensure that it works on older versions of Android as well though!

PR #629, opened 2020-11-03
Adds new convenience module using the old package name

bdemers:
Some folks are unaware of newer JJWT versions because they are still using the old io.jsonwebtoken:jjwt module. This makes automatic version updates tools "just work" for them

NOTE: This new module is _basically_ an empty jar, that adds `jjwt-api` (compile scope), `jjwt-impl` (runtime scope), and `jjwt-jackson` (runtime scope) transitive dependencies.


arvindkrishnakumar-okta (approved):

bdemers:
This will help with issues like: #695

pom.xml
```diff
@@ -99,6 +99,7 @@
         <module>api</module>
         <module>impl</module>
         <module>extensions</module>
+        <module>all</module>
```
bdemers, line 102: @lhazlewood `all` might not be the best directory name for this module, any alternative ideas? `legacy`, `convenience`, `the-module-formerly-known-as-jjwt`

**NOTE: ** the actual maven artifact name is `jjwt` but using that name for the directory may cause confusion between the project name.

lhazlewood: I updated the PR with some minor changes and changed the directory name to `tdjar` for 'transitive dependency jar' for lack of a better term.  `legacy-jar` and `convenience-jar` were too long and incongruent with the other modules.  Given that no-one except project devs will see the name, I don't think it matters that much though.

PR #647, opened 2021-02-09
Create a CI build and test using GitHub Actions.

carldea:
Utilize GitHub's CI infrastructure to build and test your project on JDKs and different OSes concurrently.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        java: [1.8]
+        java: [1.8, 11, 14]
       fail-fast: false
       max-parallel: 4
     name: Test JDK ${{ matrix.java }}, ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: Java CI build and test
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


bdemers:
Nice! Once we drop support for 1.7 (i.e. older versions of Android) we should merge this in!

bdemers (approved):

carldea:
@bdemers 
Thank you! Glad to be able to contribute to Java JSON Web Token.

skjolber:
Did you consider adding dependency caching to improve build times?

lhazlewood:
@skjolber is that possible? I haven't used GitHub Actions myself yet - but are you saying there's a way to ask GitHub to use previously-used build instances so that the Maven dependency cache is already available on the build machine before a JJWT build kicks off?  (My initial thought is that this wouldn't be allowed because it implies state between builds, which adds a significant complexity to the build infrastructure...)

I could (likely) be misunderstanding your suggestion however 😅

carldea:
> @skjolber is that possible? I haven't used GitHub Actions myself yet - but are you saying there's a way to ask GitHub to use previously-used build instances so that the Maven dependency cache is already available on the build machine before a JJWT build kicks off? (My initial thought is that this wouldn't be allowed because it implies state between builds, which adds a significant complexity to the build infrastructure...)
> 
> I could (likely) be misunderstanding your suggestion however 😅

Yes, it's possible as the GitHub infrastructure can cache maven .m2 directory and hash your pom.xmls, if they change it invalidates the cache: 
https://docs.github.com/en/actions/guides/building-and-testing-java-with-maven#caching-dependencies 

That could be another PR add this capability.

lhazlewood:
Resolved via aa17d5094e70ef628e78ba58c8122cc7a6846635 - Thank you!

Issue #235, opened 2017-07-12
Java8 jsr310 LocalDateTime 


petekaras:
Hi,
Maybe I'm missing something so apologies if this is not an issue.
I've configured my spring boot application to use jackson-datatype-jsr310 by including in pom.xml, all outgoing json is serialized correctly.
however I am using a hashmap to set the payload of the the jwt token:
```
hashMap.put(Claims.ISSUED_AT,LocalDateTime.now);
```
and this is being serialized in the Long LocalDateTimeFormat:
```
{"hour":9,"minute":53,"second":49,"nano":288000000,"dayOfYear":193,"dayOfWeek":"WEDNESDAY","month":"JULY","dayOfMonth":12,"year":2017,"monthValue":7,"chronology":{"calendarType":"iso8601","id":"ISO"}
```
I had a similar issue in another spring boot application where I was using new ObjectMapper() and not using springBoots configured ObjectMapper.
Any ideas how I could make jjwt pick up springboots ObjectMapper which would be configured correctly ?
I'm using using jjwt version 0.7.0.

Thanks in advance for your help,
Peter

lhazlewood:
Unfortunately you cannot currently specify the ObjectMapper that the JwtBuilder will use to generate the resulting JSON.  Yet. :)

We'll use this issue to track the work to allow you to do so :)

That said, would you say this issue should be classified as "Allow me to set my own ObjectMapper" ?

Or is the issue *really* "Please support Java8 Date/Time formats ?

Thoughts?

Thanks for the issue!

petekaras:
Hi Les,
Yes, I think this is really about supporting Java8 formats.
I would be happy to contribute to this project, but am short of time right now. Will let you know if I can help out in the future if you are looking for contributors ?

simdevmon:
It would be nice to have overloaded methods e.g. in `Claims` e.g.

```
public Claims setExpiration(LocalDateTime date);
public Claims setNotBefore(LocalDateTime date);
public Claims setIssuedAt(LocalDateTime date);
```

Would you accept a PR for that?

lhazlewood:
Depends on #308.

lhazlewood:
This ticket would add support for claims of type:

- LocalDateType
- Instant

Anything else?

cassiomolin:
Can we add [`OffsetDateTime`](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/time/OffsetDateTime.html) to the list?

[...]

PR #651, opened 2021-02-24
Microoptimalization: Dates ++

skjolber:
Hi, 

I'm looking into the [performance](https://github.com/skjolber/java-jwt-benchmark) of the decoder, and though I'd start with at few low-hanging fruits.

Reduce the (unnessesary) use of throwaway Date objects.
Remove `allowSkew` variable.
Use a big enough StringBuilder.

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/Clock.java
+++ b/api/src/main/java/io/jsonwebtoken/Clock.java
@@ -28,6 +28,16 @@ public interface Clock {
      * Returns the clock's current timestamp at the instant the method is invoked.
      *
      * @return the clock's current timestamp at the instant the method is invoked.
+     * @deprecated
      */
     Date now();
+
+    /**
+     * Returns the clock's current timestamp at the instant the method is invoked.
+     *
+     * @return the clock's current timestamp at the instant the method is invoked.
+     */
+
+    long millis();
+
 }
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultClock.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultClock.java
@@ -36,8 +36,12 @@ public class DefaultClock implements Clock {
      *
      * @return a new {@link Date} instance.
      */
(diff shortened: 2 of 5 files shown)


lhazlewood:
Hi @skjolber ! Thanks very much for this - how much of a change in performance does using millis instead of Date make?

I ask because most people have asked for JDK 8 DateTime support as the default way of representing timestamps, and that's on the horizon (see #308 and other linked issues).  If that's the case, does this change still make sense?

skjolber:
@lhazlewood there is not much of a difference in performance by it self, but gotta start somewher; doing unecessary stuff on the critical path should be avoided.

I guess creating Instant instead of Date is not much better. I think long as the time format is the best and most simple, at least as an internal representation. Seems in the time PRs Instant just replaces Date, so the same fix just with Instant would be ok.

Or even better: Create a comparison method for date fields, perhaps just for internal use, so that the claims value can be compared directly, i.e. with as little overhead as possible. So in other words, I'd pass in parameters "claim name" + "limit" and the new method throws an exception if the claim is not as expected.

DanFTRX:
All this does is save 1 object instantiation (The constructor for date just calls and stores System.currentTimeMillis()).  Seems kinda unnecessary.

lhazlewood:
Just a note - we'll probably need to hold off on any changes like this until we address #311, at which point the Clock interface (and others) may change.  To minimize the amount of forced implementation requirements on a version upgrade (or across multiple upgrades), I'd like to address this all at once.

skjolber:
@DanFTRX well something is better than nothing, and this PR also simplifies the logic a bit, as a bonus. But I think if I wrote this again, I'd simply extend the claims logic; adding support for time comparisons, so that no object creation needs to happen at all. 

@lhazlewood fair enough.

lhazlewood:
After some recent updating and work on the `1.0.x` branch, we're closing this in favor of the changes associated with #235 (which will replace Date usages).

PR #654, opened 2021-03-10
Remove unused field in CompressionCodecs

zaza:


zaza:
I don't think the error from travis is related to this change:
> [ERROR] Total coverage of 99.957103% did not meet target of 100.000000%

Or is it?

bdemers:
:) I think the removal of that field, removed the eager initialization of the object, which changed the coverage slightly.

zaza:
That would explain it, oh man. Since the class is deprecated and scheduled for removal I'm gonna leave it as is, and ignore the warning in my IDE.

PR #658, opened 2021-04-11
Rename parsing methods with misleading names

### LINKED ISSUE #212: Deprecate/remove the parse() method
Following the email thread *“jjwt signature validation bypass”* and a couple of issues here (#193, #90, plus several others related to [the post in Auth0's blog](https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/)):

At least some portion of JJWT users confuse the `JwtParser#parse()` to be the main method for parsing **signed** JWT-s, while in fact it will also happily parse **unsigned** JWT-s (even if `alg` header is set to something other than `none`.)

\
As a side note: I thought about proposing as a partial fix to make `parse()` throw if `alg` is not `none` and signature part is missing, but then decided against that, because it potentially could create an illusion of safety (*“no signature → failure, I tested!”*), while for the real attacker it's only one additional step to update `alg` header after stripping the signature.

Marcono1234:
Fixes #212

The interface `JwtParser` has multiple methods with misleading names:
- `parse`: Allows both signed and unsigned tokens
- `parseClaimsJwt`, `parsePlaintextJwt`: Name is misleading because JWT only describes the data format, so a signed token can also be considered a JWT. Additionally there is only a one character difference (<code>Jw<i>t</i></code> vs. <code>Jw<i>s</i></code>) between the method for unsigned and signed tokens, making it likely that the wrong method is chosen from IDE auto complete, and it is likely to be missed during code review.

This has lead to multiple applications being vulnerable due to having used the wrong method by accident.  #212 lists a few, additionally some further prominent examples (based on https://github.com/github/securitylab/issues/333):
- [DP3T](https://github.com/DP-3T/dp3t-sdk-backend/pull/208)
- [Apache Pulsar](https://github.com/apache/pulsar/pull/9172)
- [NHS Contact Tracing App](https://www.zofrex.com/blog/2020/10/20/alg-none-jwt-nhs-contact-tracing-app/)

This pull request deprecates these methods and instead adds methods with more expressive names. Note that this is a binary incompatible change because it introduces new abstract methods for the `JwtParser` interface. However, that is most likely not an issue because user code probably does not implement `JwtParser`. It appears this issue cannot be solved in a different way because this project is built with Java 7 which does not support `default` methods for interfaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ JJWT is a pure Java implementation based
 exclusively on the [JWT](https://tools.ietf.org/html/rfc7519), 
 [JWS](https://tools.ietf.org/html/rfc7515), [JWE](https://tools.ietf.org/html/rfc7516), 
 [JWK](https://tools.ietf.org/html/rfc7517) and [JWA](https://tools.ietf.org/html/rfc7518) RFC specifications and 
-open source under the terms of the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0).
+open source under the terms of the [Apache 2.0 License](https://www.apache.org/licenses/LICENSE-2.0).
 
-The library was created by [Okta's](http://www.okta.com) Senior Architect, [Les Hazlewood](https://github.com/lhazlewood)
+The library was created by [Okta's](https://www.okta.com/) Senior Architect, [Les Hazlewood](https://github.com/lhazlewood)
 and is supported and maintained by a [community](https://github.com/jwtk/jjwt/graphs/contributors) of contributors.
 
 [Okta](https://developer.okta.com/) is a complete authentication and user management API for developers.
@@ -90,7 +90,7 @@ enforcement.
  * Fully functional on all JDKs and Android
  * Automatic security best practices and assertions
  * Easy to learn and read API
- * Convenient and readable [fluent](http://en.wikipedia.org/wiki/Fluent_interface) interfaces, great for IDE auto-completion to write code quickly
+ * Convenient and readable [fluent](https://en.wikipedia.org/wiki/Fluent_interface) interfaces, great for IDE auto-completion to write code quickly
  * Fully RFC specification compliant on all implemented functionality, tested against RFC-specified test vectors
  * Stable implementation with enforced 100% test code coverage.  Literally every single method, statement and 
    conditional branch variant in the entire codebase is tested and required to pass on every build.
@@ -368,7 +368,7 @@ you more quickly and efficiently.
 <a name="quickstart"></a>
 ## Quickstart
 
-Most complexity is hidden behind a convenient and readable builder-based [fluent interface](http://en.wikipedia.org/wiki/Fluent_interface), great for relying on IDE auto-completion to write code quickly.  Here's an example:
+Most complexity is hidden behind a convenient and readable builder-based [fluent interface](https://en.wikipedia.org/wiki/Fluent_interface), great for relying on IDE auto-completion to write code quickly.  Here's an example:
 
 ```java
 import io.jsonwebtoken.Jwts;
@@ -1397,7 +1397,7 @@ Jwts.parserBuilder()
 
     .build()
 
-    .parseClaimsJwt(aJwtString)
+    .parsePlaintextUnsignedJwt(aJwtString)
 
     .getBody()
     
@@ -1639,4 +1639,4 @@ Maintained by Les Hazlewood &amp; [Okta](https://okta.com/)
 <a name="license"></a>
 ## License
 
-This project is open-source via the [Apache 2.0 License](http://www.apache.org/licenses/LICENSE-2.0).
+This project is open-source via the [Apache 2.0 License](https://www.apache.org/licenses/LICENSE-2.0).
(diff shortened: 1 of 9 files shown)


Marcono1234:
@bdemers, what is your opinion on this?

bdemers:
@Marcono1234, In general, I'm +1, but I think the methods should be structured something like:

```java
public Jwt parse(String jwt, boolean allowUnsigned);

// default to require signature
public Jwt parse(String jwt) {
  this.parse(jwt, false);  
}
```

This would be a breaking change though, but it is something we can do for the next major release.

Thoughts?

lhazlewood:
All of this has been fixed in the `jwe` branch:  in that branch, It is impossible to have malformed JWTs that bypass the required security checks.

This is because these mandatory checks were added to the JWT RFC significantly _after_ the JWT RFCs were submitted for review.  They were ratified late in the finalization process to add extra criteria for security checks.

JJWT was written before this final ratification, so that's why there are multiple `parse*` methods (to reflect the possibilities in the RFC before finalization).  Now that these things are no longer possible, the changes in the `jwe` branch ensure the same.

The `jwe` branch added the following to the `JwtParserBuilder` interface:

https://github.com/jwtk/jjwt/blob/3ba22feb0ee1655ffe0f08aac88c996f15ba11c1/api/src/main/java/io/jsonwebtoken/JwtParserBuilder.java#L44-L59

And this is checked during parsing - in any type of JWS or JWE, regardless of subtype or data - as follows (line 403 specifically):

https://github.com/jwtk/jjwt/blob/3ba22feb0ee1655ffe0f08aac88c996f15ba11c1/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java#L387-L410

And as a result already covers the functionality in this PR (I'm pretty sure at least - please confirm).  That being the case, I'd like to close this PR if possible to avoid further confusion.  Let me know if there's anything I'm missing or that's not already solved in the `jwe` branch.

Marcono1234:
Thanks for the insight, that change looks reasonable to me and would probably solve the issue with `parse`. However, it might still be worth to rename `parseClaimsJwt` and `parsePlaintextJwt` by deprecating them and adding new methods with more expressive names.
If I see it correctly the `jwe` branch is not changing their behavior, and it is not unlikely that someone confuses the JWS methods with them. (But on the other hand they should notice during testing that their code does not accept signed JWTs at all when using the unsigned JWT methods by accident.)

What do you think?

lhazlewood:
Based on the `jwe` branch change, method renames would just be for convenience since the standard `parse` logic will catch any `alg` header issue.

But I'm not sure that changing the names will matter at all - as you say, when a developer tries to use one when it should be the other, their tests (or app) will fail, at which point they'll probably read the JavaDoc or main documentation and then realize why they chose the wrong method.

There are essentially only 3 types of JWTs<sup>1</sup>:

* Signed JWTs (called a `JWS` by the RFCs)
* Encrypted JWTs (called a `JWE` by the RFCs)
* Unsecured JWTs (called an `Unsecured JWT` by the RFCs).

    <sup>1</sup> Technically there is also a concept of a Nested JWT, but it's really just any of the above three as a payload inside either a JWS or JWE.  I'm not sure if this 'counts' as a separate type.

This explains JJWT's *parse methods:  if you don't choose a *Jwe method or a *Jws method - i.e. a *Jwt method - you are by definition choosing the only thing that can possibly remain, an Unsecured JWT.

Since this is apparently not intuitive for people, the *Jwt methods could be renamed, but it's hard to know what they should be.  Consistency across all methods is ideal, but I'm not sure that's possible, because they are ideally consistent already.  There are three approaches for this IMO:

### Approach A

Don't change anything.  The `parse` method and `alg` header thing aren't issues.  People that accidentally choose the wrong method will find out during testing or production with no negative security side-effects.

### Approach B

Rename just the parse*Jwt methods to be:

`parseUnsecuredClaimsJwt` and
`parseUnsecuredPlaintextJwt`

This leaves the `parseClaimJws`, `parsePlaintextJws`, `parseClaimsJwe` and `parsePlaintextJwe` methods untouched.

### Approach C

Rename all methods to:

* `parseUnsecuredClaimsJwt`
* `parseUnsecuredBinaryJwt`
* `parseSignedClaimsJwt`
* `parseSignedBinaryJwt`
* `parseEncryptedClaimsJwt`
* `parseEncryptedBinaryJwt`

This ensures a more homogenous naming style, but has the downside of changing method names on methods that are not being misused.

But this poses a question for Nested JWTs:  even though they're not a new type of JWT, does it mean that yet more methods would need to be added?

* `parseSignedNestedJwt`
* `parseEncryptedNestedJwt`

and then what about the JWT that is nested?  Does that mean:

* `parseSignedNestedClaimsJws`
* `parseSignedNestedBinaryJws`
* `parseSignedNestedClaimsJwe`
* `parseSignedNestedBinaryJwe`
* `parseSignedUnsecuredClaimsJwt`
* `parseSignedUnsecuredBinaryJwt`
* `parseEncryptedNestedClaimsJwe`
* `parseEncryptedNestedBinaryJwe`
* ... etc ...

?

Probably not that last chunk - I'm assuming most people would agree that type of method explosion is undesirable.


----------

My current thoughts are that, since it's no longer possible to accidentally use the wrong method, I think it's probably better to stick with Approach A.  I dislike backwards-incompatible or name changes when they can be avoided without negative side-effects.  But I definitely am open to feedback from others and go with the most desired approach.

Any opinions on which approach we should take long term?

@bdemers or @dogeared (or anyone else) - want to chime in?

lhazlewood:
Closing just as a matter of housekeeping since parsing unprotected tokens is not possible by default.  I created #834 to track associated work for name changes.  Thank you @Marcono1234 !

PR #664, opened 2021-05-30
Update installation in gradle 7+

wkdalsgh192:
The `compile` and `runtime` dependency has been deprecated and finally not supported from Gradle 7+.
So, I updated the installation part in `readme.md` for the users with the latest version of Gradle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,10 @@
 JJWT aims to be the easiest to use and understand library for creating and verifying JSON Web Tokens (JWTs) on the JVM
 and Android.
 
-JJWT is a pure Java implementation based 
-exclusively on the [JWT](https://tools.ietf.org/html/rfc7519), 
-[JWS](https://tools.ietf.org/html/rfc7515), [JWE](https://tools.ietf.org/html/rfc7516), 
-[JWK](https://tools.ietf.org/html/rfc7517) and [JWA](https://tools.ietf.org/html/rfc7518) RFC specifications and 
+JJWT is a pure Java implementation based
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -302,23 +317,28 @@ dependencies {
             // Uncomment the next line if you want to use RSASSA-PSS (PS256, PS384, PS512) algorithms:
             //'org.bouncycastle:bcprov-jdk15on:1.60',
             'io.jsonwebtoken:jjwt-jackson:0.11.2' // or 'io.jsonwebtoken:jjwt-gson:0.11.2' for gson
+    // Gradle 7+ doesn't support the compile and runtime configuration anymore. Please use implementation and runtimeOnly instead.
```
bdemers, line 320: This comment is not needed as this PR covers the change already.

or maybe mention that older versions of Gradle require something different? (I’m not sure the original intent)

bdemers (changes requested):
Thanks for the PR!!

The white space and formatting changes make this diff difficult to read, can you revert that part to keep this focused the thr Gradle changes you are suggesting?

I’d also caution agains white-space/formatting PRs as they make future diff/blame difficult to follow. (If you still think it will improve the maintenance of the Readme, feel free to submit the PR!)

Thanks again!

README.md
wkdalsgh192: Thanks for the review!
Seems that the white-space/formatting changed in PRs without an intention. 
Also, I also think that the part you mentioned above is not needed. 
Will revert that part and try a new PR!

PR #665, opened 2021-05-31
Update readme

wkdalsgh192:
The white space/formatting changes were due to using prettier in vscode.
I reverted the original commit and made a new one only with proper code changes.
pls lmk if this PR has another issue. Thanks!

P.S. Do you know how to add my new commits to the closed PR and reopen it? Think it'd be better otherwise, I have to make another PR on the same Gradle issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,8 +297,8 @@ If you're building a (non-Android) JDK project, you will want to define the foll
 
 ```groovy
 dependencies {
-    compile 'io.jsonwebtoken:jjwt-api:0.11.2'
-    runtime 'io.jsonwebtoken:jjwt-impl:0.11.2',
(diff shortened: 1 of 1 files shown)


bdemers (approved):

lhazlewood:
Do we know these changes will work for all common Android Studio installations?  Or do they only work with the latest?

bdemers:
The best I could narrow it down to was Gradle deprecated these scopes at some point in the 4.x line (2017), and they were removed in the latest 7.0 release.

Android Studio (and corresponding plugin v3+) would be fine with either.

I think we can assume Android Gradle folks that are on old versions will know to swap `implementation` and `complie`.  But if we start getting questions on this we can tweak the README again 🤔

PR #671, opened 2021-07-01
Adds handling for common JSON parsing exceptions and wraps them in a `JwtException`

bdemers:
Move the parser error handling logic out of DefaultJwtParser into the new JwtDeserializer and wraps them with developer freiendly exceptions
Add check for common JSON parsing exceptions like stack overflow when parsing deeply nested (or malformed) JSON

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 This patch release:
 
-* Adds handling for common JSON parsing exceptions and wraps them in a `JwtException`.
+* Adds additional handling for rare JSON parsing exceptions and wraps them in a `JwtException` to allow the application to handle these conditions as JWT concerns.
 
 ### 0.11.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Release Notes
 
+### 0.11.3
+
+This patch release:
+
+* Adds handling for common JSON parsing exceptions and wraps them in a `JwtException`.
+
 ### 0.11.2
 
 This patch release:
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
@@ -42,7 +42,6 @@ import io.jsonwebtoken.impl.crypto.JwtSignatureValidator;
 import io.jsonwebtoken.impl.lang.LegacyServices;
 import io.jsonwebtoken.io.Decoder;
 import io.jsonwebtoken.io.Decoders;
-import io.jsonwebtoken.io.DeserializationException;
 import io.jsonwebtoken.io.Deserializer;
 import io.jsonwebtoken.lang.Assert;
 import io.jsonwebtoken.lang.DateFormats;
@@ -111,7 +110,7 @@ public class DefaultJwtParser implements JwtParser {
     @Override
     public JwtParser deserializeJsonWith(Deserializer<Map<String, ?>> deserializer) {
         Assert.notNull(deserializer, "deserializer cannot be null.");
-        this.deserializer = deserializer;
+        this.deserializer = new JwtDeserializer<>(deserializer);
         return this;
     }
 
@@ -253,7 +252,7 @@ public class DefaultJwtParser implements JwtParser {
         if (this.deserializer == null) {
             // try to find one based on the services available
             // TODO: This util class will throw a UnavailableImplementationException here to retain behavior of previous version, remove in v1.0
-            this.deserializer = LegacyServices.loadFirst(Deserializer.class);
+            this.deserializeJsonWith(LegacyServices.loadFirst(Deserializer.class));
         }
 
         Assert.hasText(jwt, "JWT String argument cannot be null or empty.");
@@ -617,11 +616,7 @@ public class DefaultJwtParser implements JwtParser {
 
     @SuppressWarnings("unchecked")
     protected Map<String, ?> readValue(String val) {
-        try {
-            byte[] bytes = val.getBytes(Strings.UTF_8);
-            return deserializer.deserialize(bytes);
-        } catch (DeserializationException e) {
-            throw new MalformedJwtException("Unable to read JSON value: " + val, e);
-        }
+        byte[] bytes = val.getBytes(Strings.UTF_8);
+        return deserializer.deserialize(bytes);
     }
 }
--- a/impl/src/main/java/io/jsonwebtoken/impl/JwtDeserializer.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/JwtDeserializer.java
@@ -24,7 +24,7 @@ import io.jsonwebtoken.lang.Assert;
 import java.nio.charset.StandardCharsets;
 
 /**
- * A {@link Deserializer} implementation that wraps another Deserializer implementation to adds common JWT related
+ * A {@link Deserializer} implementation that wraps another Deserializer implementation to add common JWT related
  * error handling.
  * @param <T> type of object to deserialize.
  * @since 0.11.3
--- a/impl/src/main/java/io/jsonwebtoken/impl/JwtDeserializer.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/JwtDeserializer.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2021 jsonwebtoken.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.jsonwebtoken.impl;
+
+import io.jsonwebtoken.MalformedJwtException;
+import io.jsonwebtoken.io.DeserializationException;
+import io.jsonwebtoken.io.Deserializer;
+import io.jsonwebtoken.io.IOException;
+import io.jsonwebtoken.lang.Assert;
+
+import java.nio.charset.StandardCharsets;
+
+/**
+ * A {@link Deserializer} implementation that wraps another Deserializer implementation to adds common JWT related
+ * error handling.
+ * @param <T> type of object to deserialize.
+ * @since 0.11.3
+ */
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Release Notes
 
+### 0.11.3
+
+This patch release:
+
+* Adds handling for common JSON parsing exceptions and wraps them in a `JwtException`.
```
lhazlewood, line 7: A minor nit: to avoid a suspicion that that we didn't handle JSON parsing exceptions previously (we did and do), I'd write this as:

> Adds additional handling for rare JSON parsing exceptions and wraps them in a \`JwtException\` to allow the application to handle these conditions as JWT concerns.

lhazlewood (changes requested):

CHANGELOG.md
bdemers: I like it!! Thanks!

bdemers, line 7:
```suggestion
* Adds additional handling for rare JSON parsing exceptions and wraps them in a `JwtException` to allow the application to handle these conditions as JWT concerns.
```

lhazlewood (approved):
Thanks!

Marcono1234 (commented):
Some minor review comments, I hope they are helpful.

impl/src/test/groovy/io/jsonwebtoken/impl/DefaultJwtParserTest.groovy
```diff
@@ -69,7 +70,8 @@ class DefaultJwtParserTest {
             }
         }
         def p = new DefaultJwtParser().deserializeJsonWith(deserializer)
-        assertSame deserializer, p.deserializer
+        assertTrue("Expected wrapping deserializer to be instance of JwtDeserializer", p.deserializer instanceof JwtDeserializer )
```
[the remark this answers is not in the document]
lhazlewood: @Marcono1234 you have a lot of good catches in this review - thanks so much for chiming in!

As for asserting deprecated methods, I see two sides to this:  1) `JwtParser#deserializeJsonWith` is deprecated, so no new functionality/patches are guaranteed for such methods, so new behavior doesn't have to be asserted, versus 2) we implemented it, so we should check it just for sanity in our own unit tests.

I'd vote in favor of approach 2 just because I like to be exhaustively thorough.  @bdemers any thoughts?  Can we add a quick test?

lhazlewood (changes requested):

impl/src/test/groovy/io/jsonwebtoken/impl/DefaultJwtParserTest.groovy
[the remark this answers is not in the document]
Marcono1234: My comment was slightly misleading. The modified test, on which this review comment is added, is already testing the behavior for the deprecated `JwtParser#deserializeJsonWith` (by calling `new DefaultJwtParser().deserializeJsonWith(deserializer)`, see code). So that is already covered by these changes.

The point was that it would be good to perform the same assertion on the non-deprecated `DefaultJwtParserBuilder` as well.

impl/src/test/groovy/io/jsonwebtoken/impl/JwtDeserializerTest.groovy
```diff
@@ -0,0 +1,76 @@
...
+        // create a Deserializer that will throw a StackOverflowError
+        Deserializer<Map<String,?>> deserializer = mock(Deserializer)
+        expect(deserializer.deserialize(jsonBytes)).andThrow(new StackOverflowError("Test exception: testParserStackOverflowError" ))
+        replay(deserializer)
+
+        try {
+            new JwtDeserializer<>(deserializer).deserialize(jsonBytes)
+            Assert.fail("Expected IOException")
+        } catch (IOException e) {
+            assertEquals JwtDeserializer.MALFORMED_COMPLEX_ERROR + new String(jsonBytes), e.message
```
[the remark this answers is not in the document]
bdemers: Good catch, I probably need to clean up line 41 which already defines the string

impl/src/test/groovy/io/jsonwebtoken/impl/JwtDeserializerTest.groovy
```diff
@@ -0,0 +1,76 @@
...
+        // create a Deserializer that will throw a DeserializationException
+        Deserializer<Map<String,?>> deserializer = mock(Deserializer)
+        expect(deserializer.deserialize(jsonBytes)).andThrow(new DeserializationException("Test exception: testDeserializationExceptionMessage" ))
+        replay(deserializer)
+
+        try {
+            new JwtDeserializer<>(deserializer).deserialize(jsonBytes)
+            Assert.fail("Expected MalformedJwtException")
+        } catch (MalformedJwtException e) {
+            assertEquals JwtDeserializer.MALFORMED_ERROR + new String(jsonBytes), e.message
```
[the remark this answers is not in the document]
bdemers: Same as above, line 62 should be cleaned up

bdemers:
@lhazlewood @Marcono1234 I think I've hit all the feedback, let me know!

lhazlewood (approved):
👍

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
```diff
@@ -617,11 +616,7 @@ protected JwtSignatureValidator createSignatureValidator(SignatureAlgorithm alg,
...
     @SuppressWarnings("unchecked")
     protected Map<String, ?> readValue(String val) {
-        try {
-            byte[] bytes = val.getBytes(Strings.UTF_8);
-            return deserializer.deserialize(bytes);
-        } catch (DeserializationException e) {
-            throw new MalformedJwtException("Unable to read JSON value: " + val, e);
-        }
+        byte[] bytes = val.getBytes(Strings.UTF_8);
+        return deserializer.deserialize(bytes);
```
Marcono1234, line 620: Does this actually compile? `deserialize` throws `DeserializationException`, which is a checked exception.

impl/src/main/java/io/jsonwebtoken/impl/JwtDeserializer.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public T deserialize(byte[] bytes) throws DeserializationException {
+        try {
+            return deserializer.deserialize(bytes);
+        } catch (DeserializationException e) {
+            throw new MalformedJwtException(MALFORMED_ERROR + new String(bytes, StandardCharsets.UTF_8), e);
+        } catch (StackOverflowError e) {
+            throw new IOException(MALFORMED_COMPLEX_ERROR + new String(bytes, StandardCharsets.UTF_8), e);
```
Marcono1234, line 51: Does this actually compile? The method does not declare in its `throws` clause that it throws the checked exception type `IOException`.

Should throw `MalformedJwtException` instead?

impl/src/test/groovy/io/jsonwebtoken/impl/DefaultJwtParserBuilderTest.groovy
```diff
@@ -92,4 +96,25 @@ class DefaultJwtParserBuilderTest {
             assertEquals DefaultJwtParserBuilder.MAX_CLOCK_SKEW_ILLEGAL_MSG, expected.message
         }
     }
+
+    @Test
+    void testDefaultDecoder() {
```
Marcono1234, line 101: Really minor, but unlike the test methods above which actually set a `Decoder`, this one sets a `Deserializer`
```suggestion
    void testDefaultDeserializer() {
```

impl/src/test/groovy/io/jsonwebtoken/impl/DefaultJwtParserBuilderTest.groovy
```diff
@@ -92,4 +96,25 @@ class DefaultJwtParserBuilderTest {
...
+        JwtParser parser = new DefaultJwtParserBuilder().build()
+        assertThat parser.jwtParser.deserializer, CoreMatchers.instanceOf(JwtDeserializer)
+
+        // TODO: When the ImmutableJwtParser replaces the default implementation this test will need updating, something like:
+        // assertThat parser.deserializer, CoreMatchers.instanceOf(JwtDeserializer)
+    }
+
+    @Test
+    void testUserSetDecoderWrapsImplementation() {
```
Marcono1234, line 110: Really minor, but unlike the test methods above which actually set a `Decoder`, this one sets a `Deserializer`
(the method name might also be misleading because not the user deserializer wraps, but it is wrapped)
```suggestion
    void testUserSetDeserializerWrapped() {
```

Marcono1234 (commented):
The changes look good, thanks!

Though, unless I am overlooking something, the Java changes would not compile due to incorrect usage of checked exceptions (see comments).

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
bdemers: `DeserializationException` is unchecked, everything compiles and tests run ;)

bdemers:
Though, it looks like Travis CI didn't respond to the PR, so I'll take a look at why that wasn't triggered

impl/src/test/groovy/io/jsonwebtoken/impl/DefaultJwtParserBuilderTest.groovy
bdemers: good catch, copypasta

impl/src/main/java/io/jsonwebtoken/impl/JwtDeserializer.java
lhazlewood: This particular `IOException` is `io.jsonwebtoken.io.IOException` - a subclass of `JwtException` specifically to represent IO issues in a runtime exception format (to ensure that callers can catch everything under the `JwtException` umbrella if they choose).  We wanted to represent this as a proper JWT `IOException` because it's not necessarily that the JSON is malformed (it could be perfectly valid, just arbitrarily complex).

bdemers:
(rebased to force trigger Travis, looks like it's getting picked up now)

impl/src/main/java/io/jsonwebtoken/impl/JwtDeserializer.java
Marcono1234: Ah you are right, sorry for the confusion! I was looking at the changes only on GitHub and confused `io.jsonwebtoken.io.IOException` with the checked exception type `java.io.IOException`.

Using `IOException` here then is reasonable.

Marcono1234 (approved):

arvindkrishnakumar-okta:
@lhazlewood Would you be able to release a patch version that contains this fix soon?

lhazlewood:
@arvindkrishnakumar-okta We can do that, probably sometime this coming week. @bdemers Thoughts?

PR #682, opened 2021-10-04
enable GitHub Actions

sullis:
This PR replaces TravisCI with GitHub Actions.

### Motivation

https://arstechnica.com/information-technology/2021/09/travis-ci-flaw-exposed-secrets-for-thousands-of-open-source-projects/


bdemers:
Thanks, @sullis! 

TL;DR After we drop support for Java 7 (e.g. v1.0 of JJWT), we can look into switching CI systems.

We need to wait on this one, JDK 7 isn't supported by this GH Action (We compile and test against 7 to ensure support with Android, and we have a few other issues/discussions on the topic)

I really like GH Actions, but the secret leak isn't unique to just Travis, various GH Actions tasks have also leaked secrets (I'm not trying to stick up for Travis CI here either)

PR #690, opened 2021-11-15
Fix an inefficient keyset iterator

TomMD:
Found via infer on Lift, an inefficient keyset iterator is in the form:

```
for key in mapping:
    entry = mapping.find(key)
```

Which is linear-log instead of the more optimal linear solution.

N.B. This commit bundles in the Lift configuration (i.e. https://github.com/marketplace/muse-dev) used when this bug was found. If you install the app then depending on the preferences, we might want to add some of the other observed issue types to an ignore list.

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/JwtMap.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/JwtMap.java
@@ -172,8 +172,9 @@ public class JwtMap implements Map<String, Object> {
         if (m == null) {
             return;
         }
-        for (String s : m.keySet()) {
-            put(s, m.get(s));
+        for (Map.Entry <? extends String, ?>entry : m.entrySet()) {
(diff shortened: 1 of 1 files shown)


bdemers:
Thanks @TomMD!
Can you split out the Lift config into a separate PR? 
I've got nothing against using Lift, but ideally, we'd configure the static analysis tools to run as part of the Maven build as opposed to only to run in CI (so they can be run during development, integrate into IDEs etc)

There is a slightly dated PR: #546 to integrate spotbugs/findsecbugs

TomMD:
@bdemers Can do.  I'll force-push this to be the fix and the config as another MR.

FYI (and disclaimer - am a Lift original cofounder) that Lift is more than FSB/SpotBugs as it also most notably includes ErrorProne and Infer which can be oh so slightly more work to get running.

PR #691, opened 2021-11-15
Add a Sonatype Lift configuration to ignore uninteresting results

TomMD:
Following up on the conversation in #690, this is a Sonatype Lift configuration if desired. It's an automation of many tools including Cobra, Semgrep, FindSecBugs, Infer, and Errorprone.  The [app is here](https://github.com/apps/sonatype-lift) and the [marketplace is there](https://github.com/marketplace/muse-dev).

### DIFF
--- a/.lift/config.toml
+++ b/.lift/config.toml
@@ -1,0 +1,4 @@
+ignoreRules = ["MissingOverride"]
+ignoreFiles = '''
+**/test/**
+'''


lhazlewood:
Thanks @TomMD !

lhazlewood:
@TomMD I have a question:

In the `jwe` branch, I've added a `.lift.toml` file to the `impl` module:

https://github.com/jwtk/jjwt/blob/414939a9e69139a24bb186fb1d9a03c5c2670734/impl/.lift.toml

But Lift isn't honoring those exclusions - I'm still getting reports for those things.  What might I be doing wrong?

TomMD:
@lhazlewood Oh, annoying UX badness. Thanks for pinging me.

*What happened*

Or what I think happened-  the code in the `impl` directory is built by more than just `cd impl ; mvn ...`  - namely, it is specified to be built by the top level pom.xml too and the top level lift config (This one in .lift/config.toml) is not set to ignore those results.

*Suggestions*

1. Don't add a second .lift.toml unless you really want to analyze a different, second, set of files compiled by running `mvn` in this sub directory.
2. Do add the extra ignore values to the `.lift/config.toml` at the top level.

*What we're going to do about it*

I've fed this annoyance to the product managers and designers. They'll figure out what preferable behavior should occur and work it into our sprints.  This doesn't mean the behavior will change soon but it is known now and we'll track the priority and progress.

lhazlewood:
Thanks for the reply, but I'm confused, based on this documentation:

https://help.sonatype.com/lift/configuration-reference#ConfigurationReference-subprojectsSubprojects

This seemed to make it obvious that my addition would work 😅 

If that's not the case, how do I add different rules then to the top level `.lift/config.toml` file when I only want them to apply to just the impl module and not the others?  Like, what would the actual syntax be?

lhazlewood:
After re-reading it, I suppose that 'subproject' != 'maven module'.  The structure looked just like a standard multi-module maven project, so I guess that's why I assumed it would work.

TomMD:
> If that's not the case, how do I add different rules then to the top level .lift/config.toml file when I only want them to apply to just the impl module and not the others? Like, what would the actual syntax be?

Right now that is an issue - you can't. The top level rules will apply to all findings from running the tools on the files compiled by the top level build system. These findings could include files in the subdirectories that have alternate configurations. In that case Lift isn't doing a smart job - applying a different rule set in those cases - and just using the rules from the directory where the build system started. I'm pushing this ugliness at the team now, hopefully everyone sees it as the same issue you do and we can get to fixing it.

lhazlewood:
Thanks for the explanation! That's helpful!  Hopefully we can enable that kind of feature in the near future - sounds super useful 😄

Issue #273, opened 2017-10-23
How do you use this in Windows?


totszwai:
Is there a way to use this in Windows? 

This is the error I'm getting when I try to pass the ProvateKey to jjwt:

`The specified key of type sun.security.mscapi.RSAPrivateKey is not an RSA PrivateKey.`

lhazlewood:
Your RSA key implementation does not implement the JDK-standard interfaces:

* `java.security.PrivateKey`
* `java.security.interfaces.RSAKey`
or:
* `java.security.interfaces.RSAPrivateKey`

Your `PrivateKey` implementation must implement these interfaces to function correctly.

It is recommended that you wrap your `sun`-specific implementation with one that implements the `java.security.interfaces.RSAPrivateKey` interface and that will cover all of your bases.  For example:

```java
public class StandardRsaPrivateKey implements java.security.interfaces.RSAPrivateKey {
    
    private final sun.security.mscapi.RSAPrivateKey target;

    public StandardRsaPrivateKey(sun.security.mscapi.RSAPrivateKey target) {
        this.target = target;
    }

    //implement the other methods and delegate to the `target` instance.  For example:
    @Override
    public BigInteger getModulus() {
        return this.target.getModulus();
    }

    //do the same for all remaining methods.
}
```

Then you can wrap your sun PrivateKey with this instance

```java
sun.security.mscapi.RSAPrivateKey platformSpecific = getWindowsPrivateKey(); //implement me

java.security.interfaces.RSAPrivateKey goodKey = 
    new StandardRsaPrivateKey(platformSpecific);
```

Then you can use the `goodKey` instance with the builder:

```java
Jwts.builder().signWith(SignatureAlgorithm.RS256, goodKey)
     // ... etc ...
     .build();
```

In the future, please ask usage questions on https://stackoverflow.com - we use GitHub issues only for actionable work on the codebase itself, not for general questions.  Thanks!

Issue #160, opened 2016-09-06
Windows RSAPrivateKey (MSCAPI)


x2on:
Hi,

im loading a RSAPrivateKey on Windows with the following code:

```java
KeyStore ks = KeyStore.getInstance("Windows-MY");
ks.load(null, null);
RSAPrivateKey privateKey = ks.getKey("someKeyAlias", null);
```

But then i get the following exception:
```java
Exception in thread "main" java.lang.IllegalArgumentException: RSA signatures must be computed using an RSAPrivateKey.  
The specified key of type sun.security.mscapi.RSAPrivateKey is not an RSAPrivateKey.
	at io.jsonwebtoken.impl.crypto.RsaSigner.<init>(RsaSigner.java:34)
	at io.jsonwebtoken.impl.crypto.DefaultSignerFactory.createSigner(DefaultSignerFactory.java:43)
	at io.jsonwebtoken.impl.crypto.DefaultJwtSigner.<init>(DefaultJwtSigner.java:37)
	at io.jsonwebtoken.impl.crypto.DefaultJwtSigner.<init>(DefaultJwtSigner.java:32)
	at io.jsonwebtoken.impl.DefaultJwtBuilder.createSigner(DefaultJwtBuilder.java:339)
	at io.jsonwebtoken.impl.DefaultJwtBuilder.compact(DefaultJwtBuilder.java:321)
```

Any idea why this is not compatible?

bhowell2:
Is this an RSA key generated with MSCAPI? Did you store this generated key in the Java Keystore? There are some articles out there that describe how to load your keys into a Java Keystore.

bhowell2:
I've also assumed in the response above that you didn't actually use `null` in the `load(null, null)` call above - i figured you would get an error thrown way before about that.

x2on:
Hi,
im loading the RSAPrivateKey from the Windows Certificate store with the code above.
It works and i get a valid key from the certificate store.
Then i use the following code:

```java
String payload = Jwts.builder()
.setPayload("somePayload")
.signWith(SignatureAlgorithm.RS256, privateKey)
.compact();
```

And this call throws the error above.

bhowell2:
[...]

PR #693, opened 2021-11-17
Fallback to checking algorithm name if key not instanceof RSAKey or EC.

netmackan:
Thank you all for a great library!

**For this pull request I am suggesting a small improvement which would take away the need to override the RsaSigner or EllipticCurveSigner to support PrivateKeys not being instances of RSAKey or ECKey but still being able to check the key type as today but to also use the method recommended in the PKCS#11 guide. The change is basically to fallback to check PrivateKey.getAlgorithm().**

Background:

PrivateKey objects protected by hardware such as smart card or HSM and using the SunPKCS11 provider do not implement RSAKey and ECKey as they are unextractable. This makes it cumbersome with a lot of overriding to do in order to use JJWT. This has previously been discussed in #273 and #160 and possibly some other all though from a different angle.

Explanation in the Java PKCS#11 guide is here:
> Key objects representing unextractable token keys should only implement the relevant generic interfaces in the java.security and javax.crypto packages (PrivateKey, PublicKey, or SecretKey). Identification of the algorithm of a key should be performed using the Key.getAlgorithm() method. 
[PKCS#11 Reference Guide section Token Keys](https://docs.oracle.com/en/java/javase/11/security/pkcs11-reference-guide1.html#GUID-508B5E3B-BF39-4E02-A1BD-523352D3AA12)

For the cases where those interfaces are used to check key parameters it is not so easy to see how this could be improved without API changes. Probably for those cases the API would have to take a KeyPair and not just the PrivateKey as the parameters could then be checked on the PublicKey object instead of the PrivateKey one. This is out of scope for this PR.

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/crypto/EllipticCurveSigner.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/crypto/EllipticCurveSigner.java
@@ -29,7 +29,7 @@ public class EllipticCurveSigner extends EllipticCurveProvider implements Signer
 
     public EllipticCurveSigner(SignatureAlgorithm alg, Key key) {
         super(alg, key);
-        if (!(key instanceof PrivateKey && key instanceof ECKey)) {
+        if (!(key instanceof PrivateKey && (key instanceof ECKey || "EC".equals(key.getAlgorithm()) || "ECDSA".equals(key.getAlgorithm())))) {
             String msg = "Elliptic Curve signatures must be computed using an EC PrivateKey.  The specified key of " +
                          "type " + key.getClass().getName() + " is not an EC PrivateKey.";
             throw new IllegalArgumentException(msg);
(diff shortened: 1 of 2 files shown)


bdemers:
+1, I took a quick shot at testing this out. After a bit of fumbling, I was able to load a private key from a Yubikey's PIV interface.
I wasn't actually able to sign anything with said key (I blame my own ignorance, cobbling together various guides).
But the resulting key pulled from the device (via a `Keystore`) was a `sun.security.pkcs11.P11Key$P11PrivateKey` (and `getAlgorithm()` returned `RSA`).

@netmackan as for the follow-up comment about the keypair, can you create a separate issue for that? @lhazlewood has been working on supporting JWE's and has been making changes in that area, (if interested check out the `jwe` branch)

bdemers (approved):

netmackan:
I submitted #700 for the possible further improvement but that would require API change. Can we have this one merged in the meantime?

bdemers:
@netmackan that sounds like a plan!

As for the other API changes
Any chance you have a minimal example that would hit all the touchpoints you mentioned in #700?

lhazlewood:
@netmackan Closing this just as a matter of housekeeping since #700 was resolved via the JWE branch work that has been merged to `master`.  If you feel that something hasn't been addressed, please do let us know and we can discuss before the upcoming release.  Thank you so much for the PR and the discussion!

Issue #700, opened 2022-01-03
API update in order to support signing through PKCS#11 (and other)


netmackan:
**Is your feature request related to a problem? Please describe.**
This issue brings up the issue that it is cumbersome to use JJWT for signing together with a smartcard/HSM as there are key parameter checks that are done on the PrivateKey instances that does not work in those cases and the user would have to make a lot of overrides and duplicate code in order to get past that.

**Describe the solution you'd like**
The suggested solution is to change/add to the API methods to get also the PublicKey and to instead use that one to make the key algorithm checks. That way the library would be possible to use for instance through PKCS#11.

**Describe alternatives you've considered**
Workaround today is to override the parts and remove or change the key parameters checks but that is very clumsy. 

**Additional context**
PrivateKey objects protected by hardware such as smart card or HSM and using the SunPKCS11 provider do not implement RSAKey and ECKey as they are unextractable. This makes it cumbersome with a lot of overriding to do in order to use JJWT. This has previously been discussed in #273 and #160 and possibly some other all though from a different angle. Recently I provided a Pull Request #693 which solved those issues that could be solved without API changes. However, that did not cover the cases where the key paramaters are being checked so some overriding is still needed by the user.

The problem is that today the parameters are checked on the private key classes but when using SunPKCS11 or other providers and the keys are protected in a smartcard or HSM those key parameters are not available in the private key instances. Instead the parameters can be obtained from the *public key* instance.

Probably for those cases the API would have to also take the PublicKey object (or to take a KeyPair object which contains both PrivateKey and PublicKey).

lhazlewood:
Not stale - still under consideration!

lhazlewood:
Just a note as I think about this:  This could impact the [jwe](#178) branch.  Implementations there currently assert `ECKey && PrivateKey` or `ECKey && PublicKey` and similar for `RSAKey` concepts via generics.  Example:

* [EllipticCurveSignatureAlgorithm](https://github.com/jwtk/jjwt/blob/3ba22feb0ee1655ffe0f08aac88c996f15ba11c1/api/src/main/java/io/jsonwebtoken/security/EllipticCurveSignatureAlgorithm.java)
* [DefaultEllipticCurveSignatureAlgorithm](https://github.com/jwtk/jjwt/blob/3ba22feb0ee1655ffe0f08aac88c996f15ba11c1/impl/src/main/java/io/jsonwebtoken/impl/security/DefaultEllipticCurveSignatureAlgorithm.java)

[...]

PR #704, opened 2022-01-20
Support pkcs11

jmcaetano:
This PR comes as part of the issue described in #700.

The purpose of this PR is to only do key length validation when we are working with public keys, otherwise we just check if the provided key is a valid private key. The reason for this is the fact that when using private keys it is not guaranteed we know the key length. Also, if the key has an invalid key length, the operation will fail anyway, however in a later stage (as java cannot perform `SHA256withECDSA` with an invalid key length).

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/Jwts.java
+++ b/api/src/main/java/io/jsonwebtoken/Jwts.java
@@ -122,9 +122,9 @@ public final class Jwts {
     }
 
     /**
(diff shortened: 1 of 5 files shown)


jmcaetano:
After thinking more on this, I believe this is not the way to go.
Probably a bigger refactor will be necessary to ensure the correct curve is actually used.

lhazlewood:
Just a quick comment:  a lot of this logic is changing places for the next release to accommodate the #178 work.  Feel free to look at that branch if you're curious.  Cheers!

PR #706, opened 2022-02-21
fix: typo in the README

jacoblindev:
Just reading through the doc trying to learn how to use it, and found a typo I think.

the resulting string **NOT is** safe to expose publicly => the resulting string **is NOT** safe to expose publicly


bdemers (approved):

bdemers:
Thanks @jacoblindev!

Issue #707, opened 2022-03-01
SecretKeySpec: SignatureAlgorithm ignored -- check broken


2019-05-10:
jjwt 0.11.2

- create a token with a SecretKeySpec:
`new SecretKeySpec(MyKeyString, SignatureAlgorithm.HS256.getJcaName())`
- check created token and the SignatureAlgorithm is `HS384` instead.

If I look at `io.jsonwebtoken.forSigningKey(Key key)` the `SignatureAlgorithm` is _completely_ ignored -- instead some length test decides what algorithm is to be used.
So, inevitably, the check in the JWTParser fails with a baffling error!

How is that even possible?
- why is an algorithm even required if it is not used?
- why is it not used?
- what's the idea in calculating an algorithm which by necessity will break the token check on the other end where the user will obviously expect the very same algorithm set when creating the token?

2019-05-10:
Actually, it's even worse:
`Error processing validation response for: ... The verification key's size is 328 bits which is not secure enough for the HS384 algorithm.`

So, obviously the Parser does not even arrive at the same algorithm as the Signer, despite using the very same key!

bdemers:
Hi @2019-05-10,

I'm not sure I'm following your description, can you add a quick snippet to help us reproduce the issue?

I took a quick pass at creating an example, but this works as expected:

```java
// populate random bytes
byte[] secretBytes = new byte[32];
new SecureRandom().nextBytes(secretBytes);

// generate key using keyspec
Key secretKey1 = new SecretKeySpec(secretBytes, SignatureAlgorithm.HS256.getJcaName());
SignatureAlgorithm algorithm1 = SignatureAlgorithm.forSigningKey(secretKey1);
System.out.println("alg for key1: " + algorithm1);

// generate key using util method
Key secretKey2 = Keys.hmacShaKeyFor(secretBytes);
SignatureAlgorithm algorithm2 = SignatureAlgorithm.forSigningKey(secretKey2);
System.out.println("alg for key2: " + algorithm2);

// show key equality
System.out.println("Keys are equal: " + secretKey1.equals(secretKey2));
```

bdemers:
@2019-05-10 ahh, thanks, i see what you mean now.

lhazlewood:
[...]

PR #709, opened 2022-03-02
Explicitly check JCA alg name when looking up SignatureAlgorithms

bdemers:
Fixes: #707

### DIFF
--- a/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
+++ b/api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
@@ -27,7 +27,10 @@ import java.security.interfaces.ECKey;
 import java.security.interfaces.RSAKey;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
 
 /**
  * Type-safe representation of standard JWT signature algorithm names as defined in the
@@ -111,6 +114,7 @@ public enum SignatureAlgorithm {
     //purposefully ordered higher to lower:
     private static final List<SignatureAlgorithm> PREFERRED_HMAC_ALGS = Collections.unmodifiableList(Arrays.asList(
         SignatureAlgorithm.HS512, SignatureAlgorithm.HS384, SignatureAlgorithm.HS256));
+    private static final Map<String, SignatureAlgorithm> PREFERRED_HMAC_ALGS_LOOKUP = preferredHmacAlgorithmMap();
     //purposefully ordered higher to lower:
     private static final List<SignatureAlgorithm> PREFERRED_EC_ALGS = Collections.unmodifiableList(Arrays.asList(
         SignatureAlgorithm.ES512, SignatureAlgorithm.ES384, SignatureAlgorithm.ES256));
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
```diff
@@ -574,8 +581,16 @@ public static SignatureAlgorithm forSigningKey(Key key) throws InvalidKeyExcepti
         if (key instanceof SecretKey) {
 
             SecretKey secretKey = (SecretKey)key;
+            String secretKeyAlg = secretKey.getAlgorithm();
             int bitLength = io.jsonwebtoken.lang.Arrays.length(secretKey.getEncoded()) * Byte.SIZE;
 
+            // first, check the key alg name
+            SignatureAlgorithm algFromJcaName = PREFERRED_HMAC_ALGS_BY_JCA.get(secretKeyAlg);
```
lhazlewood, line 588: A couple of things here:

1.  The algorithm names, as we saw on Android, can be case-insensitive.  If we're going to do a map lookup, the map keys should be lowercased (or uppercased) before map insertion to ensure a guaranteed lookup.  Additionally `secretKeyAlg` should be similarly lowercased (or uppercased) - whatever the map keys are - to ensure a valid lookup.

2.  It's not just standard JCA names that we have to worry about.  An OID String value - instead of a name - could also be returned on some JVMs, e.g. elsewhere in `SignatureAlgorithm` we have this logic:
```java
// These next checks use equalsIgnoreCase per https://github.com/jwtk/jjwt/issues/381#issuecomment-412912272
if (!HS256.jcaName.equalsIgnoreCase(alg) &&
    !HS384.jcaName.equalsIgnoreCase(alg) &&
    !HS512.jcaName.equalsIgnoreCase(alg) &&
    !HS256.pkcs12Name.equals(alg) &&
    !HS384.pkcs12Name.equals(alg) &&
    !HS512.pkcs12Name.equals(alg)) {
        throw new InvalidKeyException("The " + keyType(signing) + " key's algorithm '" + alg +
        "' does not equal a valid HmacSHA* algorithm name and cannot be used with " + name() + ".");
    }
}
```
These OID values would also have to be keys in the lookup map to ensure proper functionality.

3.  Line 585 above, `int bitLength = io.jsonwebtoken.lang.Arrays.length(secretKey.getEncoded()) * Byte.SIZE;` could throw an exception on JVMs (e.g. Hardware Security Modules) or later versions of Android because the codebase doesn't have access to the raw key bytes.  In these scenarios (`getEncoded()` throws an exception *AND* `getAlgorithm()` returns a valid identifier in the map), we should return the key as valid because there's no way for JJWT to perform a bit-strength check.  I'm wondering if we need to add this logic now however - it could be added in another PR. Probably should to isolate the two PRs in case there are any issues that arise out of it.

4. The `jwe` branch has deprecated the `SignatureAlgorithm` enum in favor of a new `SignatureAlgorithms` class because an enum doesn't allow pluggable signature algorithm implementations.  Any change in this PR needs to ensure the identical functionality exists in the `jwe` branch's `SignatureAlgorithms` class.  Can you please investigate that and open a PR there referencing the same ticket # to ensure it's captured appropriately?

lhazlewood (changes requested):

lhazlewood:
@bdemers wanna update this based on lastest master?

lhazlewood:
@bdemers bump

bdemers:
@lhazlewood rebased on master, sorry I missed the previous notification!

I do still need to deal with your comments 3&4 from above

lhazlewood:
Closing as this has been fixed as a side effect of the JWE work merged to `master`, see:

https://github.com/jwtk/jjwt/blame/26026d63cdf2a07c058695894ba1a983e1bb8cea/impl/src/main/java/io/jsonwebtoken/impl/security/DefaultMacAlgorithm.java#L97-L119

lhazlewood:
@bdemers can you please delete this branch if you're ok with that?

PR #712, opened 2022-03-15
JWE - Explicitly check JCA alg name when looking up SignatureAlgorithms

### LINKED ISSUE #113: Initial JWE support
Shared key ( `dir` ) encryption first, more complicated next.

bdemers:
Port #709 to the JWE branch


bdemers:
@dogeared I changed the target branch after creating this PR, any idea how to restart the Snyk checks?

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
```diff
@@ -654,4 +661,14 @@ public static SignatureAlgorithm forName(String value) throws SignatureException
 
         throw new SignatureException("Unsupported signature algorithm '" + value + "'");
     }
+
+    private static Map<String, SignatureAlgorithm> preferredHmacAlgorithmMap() {
+        SignatureAlgorithm[] algs = {HS256, HS384, HS512};
+        Map<String, SignatureAlgorithm> lookupMap = new HashMap<>(6);
+        for (SignatureAlgorithm alg : algs) {
+            lookupMap.put(alg.jcaName.toUpperCase(Locale.ENGLISH), alg);
+            lookupMap.put(alg.pkcs12Name.toUpperCase(Locale.ENGLISH), alg);
```
lhazlewood, line 670: FWIW, the pkcs12 name is dotted numerals - no alphabetic characters, so changing case isn't strictly necessary. ;). But leaving it the way you have it is probably a little more obvious/self-documenting and could theoretically handle any other names.

lhazlewood: Out of curiosity - is this concept of acceptable HMAC names used anywhere else given the `SignatureAlgorithm` API change in the `jwe` branch?  I'd like to ensure it exists in one place only to ensure we don't duplicate logic that might accidentally become out of sync.

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
```diff
@@ -113,6 +116,7 @@
     //purposefully ordered higher to lower:
     private static final List<SignatureAlgorithm> PREFERRED_HMAC_ALGS = Collections.unmodifiableList(Arrays.asList(
         SignatureAlgorithm.HS512, SignatureAlgorithm.HS384, SignatureAlgorithm.HS256));
+    private static final Map<String, SignatureAlgorithm> PREFERRED_HMAC_ALGS_LOOKUP = preferredHmacAlgorithmMap();
```
lhazlewood, line 119: Nit: Suggested name of `HMAC_ALGS_BY_JCA_ID` (and corresponding helper method)?  I tend to name lookup maps in this way to self-document the map keys (since 'lookup' tends to be generic').

Also, unless I'm missing something, we can drop the `PREFERRED` prefix because nothing about the map is used for preferred ordering checks, no?

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
```diff
@@ -576,8 +575,16 @@ public static SignatureAlgorithm forSigningKey(Key key) throws InvalidKeyExcepti
         if (key instanceof SecretKey) {
 
             SecretKey secretKey = (SecretKey)key;
+            String secretKeyAlg = secretKey.getAlgorithm();
```
lhazlewood, line 578: Might want an assertion here to ensure that `secretKey.getAlgorithm()` always returns a non-null/non-empty String value, since without it, the added line 582 below would cause a null-pointer exception.

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
```diff
@@ -576,8 +575,16 @@ public static SignatureAlgorithm forSigningKey(Key key) throws InvalidKeyExcepti
...
             int bitLength = io.jsonwebtoken.lang.Arrays.length(secretKey.getEncoded()) * Byte.SIZE;
 
+            // first, check the key alg name
+            SignatureAlgorithm algFromJcaName = PREFERRED_HMAC_ALGS_LOOKUP.get(secretKeyAlg.toUpperCase(Locale.ENGLISH));
+            if (algFromJcaName != null && bitLength >= algFromJcaName.minKeyLength) {
```
lhazlewood, line 583: I'm not quite sure this is correct.

Only if `bitLength` is `0` *and* you found a `SignatureAlgorithm` from the JCA ID lookup, do you want to short circuit and return the discovered `SignatureAlgorithm` instance.

Otherwise you want to fall back to the `PREFERRED_HMAC_ALGS` priority-based lookup to retain existing behavior.

If we don't want to do it this way, we will probably have to change the JavaDoc to reflect different semantics, and we'd need to think through the repercussions of different semantics.

That is, if someone calls it today and it returns alg A, and they call it after a new release and it returns alg B, that probably wouldn't be a good thing.

api/src/main/java/io/jsonwebtoken/SignatureAlgorithm.java
bdemers: Agreed! I renamed that var three times locally trying to figure out a better name 😆

lhazlewood:
Closing as a matter of housekeeping.  `master`'s `DefaultMacAlgorithm` does the appropriate name lookups (with OID support), and the old `SignatureAlgorithm` concept is fully deprecated and not used anywhere other than as a bridge to the new SignatureAlgorithm and MacAlgorithm implementations.

PR #720, opened 2022-04-15
[#719] JSON-B extension.

### LINKED ISSUE #719: Add JSON-B extension
**Is your feature request related to a problem? Please describe.**

There are gson and jackson and org.json extensions.
Let’s add one for JSON-B, so there is no need to ship a specific library.
This is useful for Java/Jakarta EE compatible servers where you do not know the implementation beforehand, but you won't need to ship one with your app either.

**Describe the solution you'd like**

Similar to this one:
https://github.com/bmarwell/shiro-jwt-showcase/blob/3e4ec3691da3a0a7a653b4b915c3532c8fb742ed/finish/src/main/java/io/github/bmarwell/shiro/jwt/json/JsonbJwtDeserializer.java

(yes, PR ahead!)

**Describe alternatives you've considered**

* Creating above class every single time for my apps
* Create a wrapper for Johnzon

**Additional context**

./.

bmarwell:
Fixes #719:

This commit will add a JSON-B extension based on [JSON-B specification](https://github.com/eclipse-ee4j/jsonb-api) 1.0.2.

## Noteworthy

* Since this specification uses Java 8 features (static methods in interfaces), the property `jdk.version` was set to `8`.
* Since we now use Java 8, there is little advantage in using custom assertions, this `null` input is validated by `java.util.Objcets.requireNonNull` and will throw a `NullPointerException` (which makes more sense anyway) instead of an `IllegalArgumentException`.
* Instead of `Strings.UTF_8` the JDK8-included `StandardCharsets.UTF_8` was used.
* I used Johnzon for testing, but only because it is an Apache project. It should work with Eclipse Yasson without any changes. If desirable, I could implement a maven profile for this – or even better, an IT project (e.g. `intergation-tests/json/jsonb-yasson` and `intergation-tests/json/jsonb-sth-else`).
* The (`provided`) dependency on json-api (aka json-p for JSON-Parser) is needed at least for Johnzon, but probably for some Exceptions as well. Johnzon fails because `javax.json.JsonException` is missing. But since json-api/json-p is a transitive dependency for json-bind-api (json-b) as well, this should be okay.

## Other notes

* The `provided` scope is because of JakarteEE Application Server usage. Tested on [OpenLiberty](https://openliberty.io/). But for JavaSE apps, users of this lib would need to include those by hand, as well as the implementation. This should go into the README.

## Tasks

- [X] Added implementation.
- [X] README.md updated.
- [X] Added notice about JavaSE vs JakartaEE environments and their ([not-]provided) dependencies.
- [X] Added "when-to-use" note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1287,8 +1287,7 @@ They are checked in order, and the first one found is used:
 Jackson will force a sizable (> 1 MB) dependency to an Android application thus increasing the app download size for 
 mobile users.
 
-If you want to use POJOs and a JSON-B compliant specification _**or**_ you want to use JJWT on a JakartaEE compliant application server, use
-`io.jsonwebtoken:jjwt-jsonb`.
+If you want to use POJOs and a JSON-B compliant specification _**or**_ you want to use JJWT on a JakartaEE compliant application server, use `io.jsonwebtoken:jjwt-jsonb`.
 
 <a name="json-custom"></a>
 ### Custom JSON Processor
--- a/README.md
+++ b/README.md
@@ -1269,10 +1269,26 @@ They are checked in order, and the first one found is used:
    Android applications _unless_ you want to use POJOs as claims.  The `org.json` library supports simple 
    Object-to-JSON marshaling, but it *does not* support JSON-to-Object unmarshalling.
 
+4. JSON-B: This will automatically be used if you specify `io.jsonwebtoken:jjwt-jsonb` as a project runtime dependency.
+   JSON-B also supports POJOs as claims with full marshaling/unmarshaling as necessary.
+
+   **NOTE**: `JSON-B` is just a specification and does not bring an implementation.
+   
+   * In Java-SE environments you will need to add three more dependencies:
+      1. `jakarta.json:jakarta.json-api`
+      2. `jakarta.json.bind:jakarta.json.bind-api`
+      3. A JSON-B compliant implementation like [Eclipse Yasson](https://github.com/eclipse-ee4j/yasson) or [Apache Johnzon](https://johnzon.apache.org/johnzon-jsonb/index.html).
+   * In Java/Jakarta EE environments, you might need to enable the following features:
+      1. json-p / json-api
+      2. json-b / json-bind
+
 **If you want to use POJOs as claim values, use either the `io.jsonwebtoken:jjwt-jackson` or 
 `io.jsonwebtoken:jjwt-gson` dependency** (or implement your own Serializer and Deserializer if desired). **But beware**, 
 Jackson will force a sizable (> 1 MB) dependency to an Android application thus increasing the app download size for 
 mobile users.
+
+If you want to use POJOs and a JSON-B compliant specification _**or**_ you want to use JJWT on a JakartaEE compliant application server, use
+`io.jsonwebtoken:jjwt-jsonb`.
 
 <a name="json-custom"></a>
 ### Custom JSON Processor
--- a/extensions/jsonb/src/main/java/io/jsonwebtoken/jsonb/io/JsonbDeserializer.java
+++ b/extensions/jsonb/src/main/java/io/jsonwebtoken/jsonb/io/JsonbDeserializer.java
@@ -25,7 +25,7 @@ import java.nio.charset.StandardCharsets;
 import static java.util.Objects.requireNonNull;
 
 /**
- * @since 0.10.0
+ * @since JJWT_RELEASE_VERSION
  */
 public class JsonbDeserializer<T> implements Deserializer<T> {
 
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -1269,11 +1269,27 @@ They are checked in order, and the first one found is used:
...
+   * In Java-SE environments you will need to add three more dependencies:
+      1. `jakarta.json:jakarta.json-api`
+      2. `jakarta.json.bind:jakarta.json.bind-api`
+      3. A JSON-B compliant implementation like [Eclipse Yasson](https://github.com/eclipse-ee4j/yasson) or [Apache Johnzon](https://johnzon.apache.org/johnzon-jsonb/index.html).
+   * In Java/Jakarta EE environments, you might need to enable the following features:
+      1. json-p / json-api
+      2. json-b / json-bind
+
 **If you want to use POJOs as claim values, use either the `io.jsonwebtoken:jjwt-jackson` or 
 `io.jsonwebtoken:jjwt-gson` dependency** (or implement your own Serializer and Deserializer if desired). **But beware**,
```
bdemers, line 1286:
```suggestion
**If you want to use POJOs as claim values, use either the `io.jsonwebtoken:jjwt-jackson`, `io.jsonwebtoken:jjwt-jsonb`, or
`io.jsonwebtoken:jjwt-gson` dependency** (or implement your own Serializer and Deserializer if desired). **But beware**, 
```

README.md
```diff
@@ -1269,11 +1269,27 @@ They are checked in order, and the first one found is used:
...
+      2. json-b / json-bind
+
 **If you want to use POJOs as claim values, use either the `io.jsonwebtoken:jjwt-jackson` or 
 `io.jsonwebtoken:jjwt-gson` dependency** (or implement your own Serializer and Deserializer if desired). **But beware**, 
 Jackson will force a sizable (> 1 MB) dependency to an Android application thus increasing the app download size for 
 mobile users.
 
+If you want to use POJOs and a JSON-B compliant specification _**or**_ you want to use JJWT on a JakartaEE compliant application server, use
+`io.jsonwebtoken:jjwt-jsonb`.
+
```
bdemers, line 1292:
```suggestion
If you want to use POJOs and a JSON-B compliant specification _**or**_ you want to use JJWT on a JakartaEE compliant application server, use `io.jsonwebtoken:jjwt-jsonb`.

```

extensions/jsonb/src/main/java/io/jsonwebtoken/jsonb/io/JsonbDeserializer.java
```diff
@@ -0,0 +1,66 @@
...
+import io.jsonwebtoken.io.Deserializer;
+
+import javax.json.bind.Jsonb;
+import javax.json.bind.JsonbException;
+import java.nio.charset.StandardCharsets;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * @since 0.10.0
```
bdemers, line 28: This would need to be 0.12 or higher.

Actually, you can use `@since JJWT_RELEASE_VERSION` and we will set the version before the release 😄

extensions/jsonb/src/main/java/io/jsonwebtoken/jsonb/io/JsonbSerializer.java
```diff
@@ -0,0 +1,75 @@
...
+
+import javax.json.bind.Jsonb;
+import javax.json.bind.JsonbBuilder;
+import javax.json.bind.JsonbException;
+import java.nio.charset.StandardCharsets;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * @since 0.10.0
```
bdemers, line 31: fix version

extensions/pom.xml
```diff
@@ -37,5 +37,6 @@
         <module>jackson</module>
         <module>orgjson</module>
         <module>gson</module>
+        <module>jsonb</module>
```
bdemers, line 40: This is going to fail on the Java 7 build.
We are currently compiling for 7 for legacy (Android) support.

This can probably be pushed into a profile for java 1.8+ though

bdemers:
Great work!

README.md
bmarwell: Uhm, you were changing a line which I didn’t touch. Did you mean to do that? 
I did not implement the `Map` constructor for jsonb (nor did any of you for gson)...

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 This patch release:
 
 * Adds additional handling for rare JSON parsing exceptions and wraps them in a `JwtException` to allow the application to handle these conditions as JWT concerns.
-* Upgrades the `jjwt-jackson` module's Jackson dependency to `2.9.10.7`.
+* Upgrades the `jjwt-jackson` module's Jackson dependency to `2.12.6.1`.
```
bdemers, line 8: You might need to rebase, I merged a PR with this yesterday

bmarwell:
Oh, shouldn’t the profile name be "nonJDK7"? And did I, by accident, not rebase the PR?

bdemers:
good point, java8plus, nonJDK7 or something like that would work

bmarwell:
> good point, java8plus, nonJDK7 or something like that would work

Nah you already have "nonJDK7" in your root-`pom.xml`.
I will rebase this PR later when I have a first version of my modules-PR available. I finished it yesterday evening but didn’t push it yet, just a moment.

Issue #678, opened 2021-08-17
Jwts.builder() - setClaims - overwrite others propertyes


schotten:
If the form do you make the Jwts.builder(), the setClaims overwrite the Subject in the wrong form.
I will sent do exemple:

**Firstly i create the claims and after i will create the JWT token string**

> public String generateProfileToken() {
>         Map<String, Object> claims = new HashMap<>();
>         claims.put("aut", "anything");
>         claims.put("company",  "anything");
>         claims.put("department",  "anything");
>         return generateIdentificationToken(claims, "my login");
>     }

**The create JWT token to form that the Subject is acessyble in future**

> private String generateIdentificationToken(Map<String, Object> claims, String username) {
>         return Jwts.builder()
>                 .setClaims(claims)
>                 .setSubject(username)
>                 .setIssuedAt(new Date(System.currentTimeMillis()))
>                 .setExpiration(new Date(System.currentTimeMillis() + TOKEN_EXPIRED))
>                 .signWith(SignatureAlgorithm.HS256, secret)
>                 .compact();
>     }

**The read JWT token and get the login**

> public String getUserByToken(String jwtToken){
>         return Jwts.parser()
>                 .setSigningKey(secret)
>                 .parseClaimsJws(jwtToken)
>                 .getBody()
>                 .getSubject();
>     }

**The error:**
If i change the order for builder in function generateIdentificationToken, the subject not can accesible in function getUserByToken.
The incorret form: (I seted the subject before the claims). In this case, the subject not acessible in the future.
> private String generateIdentificationToken(Map<String, Object> claims, String username) {
>         return Jwts.builder()
>                 .setSubject(username)
>                 .setClaims(claims)
>                 .setIssuedAt(new Date(System.currentTimeMillis()))
>                 .setExpiration(new Date(System.currentTimeMillis() + TOKEN_EXPIRED))
>                 .signWith(SignatureAlgorithm.HS256, secret)
>                 .compact();
>     }

for using a builder, is it a bug in my opinion. The builder is expansive and shouldn't overwrite others parameters the token wuen the setClaims is executed.

I record the problem and attachment the video in below link:

https://user-images.githubusercontent.com/57331623/129746155-009f48ee-8ae8-4408-bb7c-d9ed268f5ee7.mp4

bdemers:
Hi @schotten!

[...]

PR #721, opened 2022-04-16
[#687] trim payload before try to parse it.

### LINKED ISSUE #687: JWT with JSON body seen as plaintext JWT
Hi,

We are facing a this behavior when using the latest release of jjwt (`0.9.1`):

When parsing the claims of a token that has a JSON body using `Jwts.parser().[...].parseClaimsJws(token)`, the token is seen as a plaintext JWT. Here is what the token contains:

Header:
```json
{
  "alg": "HS256",
  "typ": "JWT"
}
```

Payload
```json
{
  "iss": "XXX",
  "scope": "XXX",
  "onBehalfOf": "",
  "aud": " ",
  "locale": "",
  "iat": 1636109922773,
  "exp": 1636110522773,
  "jti": "ID"
}
```

Apparently, the token payload contains an extra \r\n at the end, which makes it interpreted as a plaintext jwt, see this line: https://github.com/jwtk/jjwt/blob/3e6c9e978ccb0eccf5ff83cfb04b052043839d4f/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java#L333

Is it a normal behavior ? Should the JWT be seen as plaintext when it doesn't end exactly with } ?

Thank you !

bmarwell:
fixes #687 

Hint: I really don't like your heuristics.
Imagine this scenario:

You have a binary payload which is by design not intended to be json, e.g. HOCON (typesafe config).
The parser won't parse it as either JSON nor as payload just because it starts and ends with `{` and `}`.

Better approach:
* Try to parse as JSON no matter what
* If it fails, just log info message and use the payload as-is.

If you are interested in this approach, I can create another PR. :)

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
@@ -42,6 +42,7 @@ import io.jsonwebtoken.impl.crypto.JwtSignatureValidator;
 import io.jsonwebtoken.impl.lang.LegacyServices;
 import io.jsonwebtoken.io.Decoder;
 import io.jsonwebtoken.io.Decoders;
+import io.jsonwebtoken.io.DeserializationException;
 import io.jsonwebtoken.io.Deserializer;
 import io.jsonwebtoken.lang.Assert;
 import io.jsonwebtoken.lang.DateFormats;
@@ -330,7 +331,7 @@ public class DefaultJwtParser implements JwtParser {
 
         Claims claims = null;
 
-        if (!payload.isEmpty() && payload.charAt(0) == '{' && payload.charAt(payload.length() - 1) == '}') { //likely to be json, parse it:
+        if (!payload.isEmpty() && payload.trim().charAt(0) == '{' && payload.trim().charAt(payload.length() - 1) == '}') { //likely to be json, parse it:
             Map<String, Object> claimsMap = (Map<String, Object>) readValue(payload);
             claims = new DefaultClaims(claimsMap);
         }
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
@@ -331,7 +331,8 @@ public class DefaultJwtParser implements JwtParser {
 
         Claims claims = null;
 
-        if (!payload.isEmpty() && payload.trim().charAt(0) == '{' && payload.trim().charAt(payload.length() - 1) == '}') { //likely to be json, parse it:
+        final String trimmedPayload = payload.trim();
+        if (!payload.isEmpty() && trimmedPayload.charAt(0) == '{' && trimmedPayload.charAt(payload.length() - 1) == '}') { //likely to be json, parse it:
             Map<String, Object> claimsMap = (Map<String, Object>) readValue(payload);
             claims = new DefaultClaims(claimsMap);
         }


bdemers (commented):
I agree, the checking for the leading `{` is a pain.
We have talked about this a few times as @lhazlewood has been working away on the "jwe" branch. 

IIRC, we talked about using the `cty` header, for this.

The `cty` header:
> It will typically not be used by applications when the kind of object is already known.  This parameter is ignored by JWS implementations; any processing of this parameter is performed by the JWS application.  Use of this Header Parameter is OPTIONAL.

Per spec, the JWT payload could be anything, though its _usually_ JSON, so the default should assume JSON, and anything else should use a `parseBytes()` method.

This came up recently too, this method isn't 100% correct, it _should_ return a byte[] (well, and have it's name tweaked)
https://github.com/jwtk/jjwt/blob/72973f9b9ba72438d97834db2bd56e36d3c48592/api/src/main/java/io/jsonwebtoken/JwtParser.java#L506

If interested in hacking on this send a PR to the `jwe` branch 😄

bdemers:
Also, thanks @bmarwell!

lhazlewood:
The concept of 'plaintext' (as a String) will be going away for 1.0. JJWT's concept of String vs JSON body support was created when the JWT RFCs weren't yet finalized.  Now that they are final, the reality is that a JWT can contain *any* type of payload, with JSON and JSON claims being the most common type of payload.

This is very cool IMO, and not many folks are aware of this super cool capability.  It means that JWTs can be used to represent images, pdf documents, etc - basically anything.  JWT is now effectively a general-purpose messaging format, not something limited to just identity assertions.

As such, if it's not a commonly-expected JSON claims payload, the `cty` header _should_ be set, and the application developer will be able to get that body as a `byte[]` and do what they want with it.

As @bdemers quoted above, if the `cty` header param is set, then the JWT RFC spec is clear that a JWT library like JJWT must ignore the body/payload, because it is expected that the application developer will handle the payload themselves.  As a result, the `jwe` branch currently has the heuristic written as such:

https://github.com/jwtk/jjwt/blob/3ba22feb0ee1655ffe0f08aac88c996f15ba11c1/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java#L511-L514

(but without the raw `byte[]` array stuff yet - that will be handled in subsequent work before releasing 1.0).

Now - with regards to trimming the resulting String before employing any heuristics:

Because the payload can be anything - even a normal string - we can't alter the payload for trying to detect if it's JSON (because that leading whitespace could be an expected part of a non-JSON payload).  If anything, the best thing to do is extract this logic out to a helper method, e.g.

```java
private static boolean isLikelyJson(String payload) {
   // 1. simple null/empty check:
    if (Strings.isEmpty(payload) {
        return false;
    }
    //2. Start at the beginning of the string, and iterate character by character, stopping at the first non-whitespace character
    //     if, a `{`, continue, if not, return `false`

    //3.  Start at the end of the string and iterate character by character, stopping at the first non-whitespace character
    //.    if a `}`, return `true`, if not, return `false`
}
```

Something to that effect.  This ensures we don't create another non-mutable String in the String table, and doesn't alter the intended payload.

bmarwell:
> Now - with regards to trimming the resulting String before employing any heuristics:
> 
> Because the payload can be anything - even a normal string - we can't alter the payload for trying to detect if it's JSON (because that leading whitespace could be an expected part of a non-JSON payload). If anything, the best thing to do is extract this logic out to a helper method, e.g.
> 
> ```java
> private static boolean isLikelyJson(String payload, Header<?> header) {
>    // 1. simple null/empty check:
>     if (Strings.isEmpty(payload) {
>         return false;
>     }
>     // 2. if (hasContentType(header), return `false`.
> 
>     // 3. Start at the beginning of the string, and iterate character by character, stopping at the first non-whitespace character
>     //     if, a `{`, continue, if not, return `false`
> 
>     // 4.  Start at the end of the string and iterate character by character, stopping at the first non-whitespace character
>     //.    if a `}`, return `true`, if not, return `false`
> }
> ```
> 
> Something to that effect. This ensures we don't create another non-mutable String in the String table, and doesn't alter the intended payload.

Even after reading this, I still don't get why just trying to parse it is not an option. Besides, my option does not alter the input...

If you are just trying to parse JSON and add an try-except, the parser will take care of superfluous whitespace.

Anyway, I was trying to get a quick fix into the current version. Your branch might take another month or so to evolve. So why not start with an 80% solution?

Long story short:
This is a serious but in jjwt. User delivers perfectly fine JSON and we reject it because of choosing heuristics instead of using a parser.

lhazlewood:
> Even after reading this, I still don't get why just trying to parse it is not an option. Besides, my option does not alter the input...

It doesn't alter the input directly, but it does add a potentially-large temporary `String` to the heap unnecessarily (the JWT payload is often the largest String in the JWT token in most production applications).  Given how many hundreds or thousands of JWTs that can be parsed per second in many apps (every single http request, JMS or MQTT message, etc), it's a performance hit that I wouldn't want to impose on application developers.  Even though the JVM is pretty efficient, if we can avoid the hit to the heap, all the better.

> If you are just trying to parse JSON and add an try-except, the parser will take care of superfluous whitespace.

Invoking a JSON parser and swallowing a subsequent exception when the payload isn't JSON is a heavyweight operation that could significantly impact performance - better to avoid it if possible.  I've also never really been a fan of using try/catch scenarios for expected logic, because it's not 'exceptional' any longer.  But that's less of a concern for me - performance is more important IMO.

But even if we ignore the heap and potential performance impact, there's a bigger problem here: How do you distinguish between JSON-but-invalid-JSON and not-JSON-but-should-be-swallowed/ignored exceptions?

They need to be handled differently, and there's no way to know really without inspecting the `cty` header, and that's not being done today.  And unfortunately we probably can't add the `cty` check until the next major release (i.e. 1.0), because it fundamentally changes the semantics of how JJWT works today: if anyone is setting that header today, and they are still expecting JJWT to parse the body as `Claims`, JJWT would stop working for them as they expect.  This kind of change should ideally only be done on a major release IMO.

> Anyway, I was trying to get a quick fix into the current version. Your branch might take another month or so to evolve. So why not start with an 80% solution?

We can do that today.  The quick fix is to create the helper method `isLikelyJson` as suggested above, but *not* check the `cty` header as I proposed above. (i.e. do steps 1, 3, and 4).

This ensures there is zero heap or performance penalty, and retains current behavior semantics until we can enforce the `cty` check for the 1.0 release.

> This is a serious but in jjwt. User delivers perfectly fine JSON and we reject it because of choosing heuristics instead of using a parser.

I agree we should be more tolerant here per the Robustness Principle, but in fairness, I wouldn't exactly call it serious.  To the best of my knowledge, this is the first report I can remember in many years of leading whitespace being an issue, probably because practically every other JWT library doesn't add extraneous whitespace.

That said, we do need address this, and I think(?) the minimal 3-part `isLikelyJson` suggestion above will solve all concerns until we can add the `cty` check to it for 1.0.

Does that work?

bmarwell:
> Does that work?

Phew, I don’t really know… for several reasons.

> It doesn't alter the input directly, but it does add a potentially-large temporary `String` to the heap unnecessarily (the JWT payload is often the largest String in the JWT token in most production applications). Given how many hundreds or thousands of JWTs that can be parsed per second in many apps (every single http request, JMS or MQTT message, etc), it's a performance hit that I wouldn't want to impose on application developers. Even though the JVM is pretty efficient, if we can avoid the hit to the heap, all the better.

That's a potential +1 for the Json parser, btw. ;-) 

> > If you are just trying to parse JSON and add an try-except, the parser will take care of superfluous whitespace.
> Invoking a JSON parser and swallowing a subsequent exception when the payload isn't JSON is a heavyweight operation that could significantly impact performance - better to avoid it if possible.

Is it? They can operate on streams and fail early, especially if you tell them to deserialize a  `JsonObject`. 
The instance of the JSON parser is somewhere in-memory already, probably.
If you'd really want performance optimizations, they should use a (ByteArray)InputStream or something.

That said, the more effort we put into heuristics or any other `isLikelyJson` method, the lesser the gap to parsing and try-catching an exception.

> I've also never really been a fan of using try/catch scenarios for expected logic, because it's not 'exceptional' any longer. But that's less of a concern for me - performance is more important IMO.

I totally agree here.

> But even if we ignore the heap and potential performance impact, there's a bigger problem here: How do you distinguish between JSON-but-invalid-JSON and not-JSON-but-should-be-swallowed/ignored exceptions?

At the moment we have an even bigger problem described in #678 by the original author: 
Not sure how much we want to blow up the `isLikelyJson()` method, but there is a good chance that we miss valid json and give the user a String/`byte[]` payload when we should not.

At least that is a bigger problem for me, because jjwt is already using JSON parsers which would solve the problem. 
As you mentioned: Yes, there is a cost. But how high is that cost compared to delivering a String instead of Claims at (seemingly) random?

The workaround would be for users to implement the callback parseJws/parseJwt methods where THEY do the try-catch logic. Not very user friendly when the library could do this for you.


----

I fully agree with all the following cty stuff, but I am still looking for a fix for a minor version.

 > They need to be handled differently, and there's no way to know really without inspecting the `cty` header, and that's not being done today. And unfortunately we probably can't add the `cty` check until the next major release (i.e. 1.0), because it fundamentally changes the semantics of how JJWT works today: if anyone is setting that header today, and they are still expecting JJWT to parse the body as `Claims`, JJWT would stop working for them as they expect. This kind of change should ideally only be done on a major release IMO.
> 

----



> > Anyway, I was trying to get a quick fix into the current version. Your branch might take another month or so to evolve. So why not start with an 80% solution?
> 
> We can do that today. The quick fix is to create the helper method `isLikelyJson` as suggested above, but _not_ check the `cty` header as I proposed above. (i.e. do steps 1, 3, and 4).
> 
> This ensures there is zero heap or performance penalty, and retains current behavior semantics until we can enforce the `cty` check for the 1.0 release.

As mentioned above, there is a high potential that we are not giving back a ClaimJWT/JWS but instead a byte[] payload when the JWT contained perfectly valid JSON.
Unless you want to write a __really__ fancy `isLikelyJson` method.
… which will undermine your performance assumptions, because the JSON libs already do this for you (and can probably do it better).



> That said, we do need address this, and I think(?) the minimal 3-part `isLikelyJson` suggestion above will solve all concerns until we can add the `cty` check to it for 1.0.

🤷🏻‍♂️ 
-0.5 for `isLikelyJson` because either it gets very big and complicated so there is no performance saving, or it will miss some non-edge-cases.

I mean, we *could* do a performance measurement.

lhazlewood:
> > Invoking a JSON parser and swallowing a subsequent exception when the payload isn't JSON is a heavyweight operation that could significantly impact performance - better to avoid it if possible.
> 
> Is it?

I'm fairly confident that trying logic when not needed, throwing an exception, filling in a stack trace, propagating the call stack, etc, is most definitely slower and more cumbersome than just skipping whitespace in a loop, especially when 99.99% of payloads never have leading or trailing whitespace.

> That said, the more effort we put into heuristics or any other `isLikelyJson` method, the lesser the gap to parsing and try-catching an exception.

I'm not worried about this particular case because the fix is trivial:  We'd be keeping logic that has been working for 7+ years, and just accounting for any leading or trailing whitespace that might occur.  This is what the JWT RFC requires for JSON Claims anyway.  Once that logic exists, I'm pretty sure it won't need to change for another 7 years (if ever).

We can go straight to the 'if the `cty` header is missing then parse blindly' approach for the 1.0 release and remove all heuristics entirely, but I'm unwilling to make that change now because that would actually change behavior for app developers, and we can't do that until 1.0.

So in summary, skipping leading and trailing whitespace in the payload before checking for a `{` and `}` 1) requires the least amount of change or risk to existing JJWT users and 2) fully resolves #687.  I'm not comfortable with the parse and try/catch approach until 1.0.

I'm really not trying to be difficult, I promise!  I like the parse/try/catch approach, but I want to do the bare minimum necessary to resolve #687 without adding any additional risk whatsoever.

bmarwell:
Ok, I'll try to come up with something new.

lhazlewood:
Closing this just as a matter of housekeeping: the corresponding logic for this is in `master` (after #279 was merged), here: https://github.com/jwtk/jjwt/blob/fa1e32bf8633f8935814f6d77b1257ba601402ae/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java#L257-L325

PR #722, opened 2022-04-20
[#519] Add moditect-generated module-info.java files and a simple IT.

### LINKED ISSUE #519: Add module-info
I'm trying to use modules in a support library, but there seems to be no module-info for jjwt. I suggest following either Jackson's or SLF4J's approach.

bmarwell:
Fixes #519


bmarwell:
TBD: I am hitting this bug: https://github.com/moditect/moditect/issues/62
So once I extracted the module info parts, I can switch to using `moduleInfoSource` instead of `moduleInfo`, which should make it work on Java 8 (and possibly 7).

bmarwell:
Did a quick test:
`./mvnw install` on JDK 8, then `./mvnw verify -pl integration-tests/unsigned-jackson` on JDK11 which passed.
This should probably be added to the CI.

lhazlewood:
Just a quick note, and this is probably obvious, but this can't be merged until after a 1.0 release due to the requirement for JDK 8.

bmarwell:
@lhazlewood this only requires Java 8 at RELEASE time, not at runtime.

Are you sure this is what you intended to say? 🤔

Talked to @bdemers the other day about it. We were thinking of a release profile which included an enforcer check for JDK8.
That would produce JDK7 code in all cases.

I really don't see why this needs to go into the next major version.

bmarwell:
Also, you might just have read past my comments mentioning 1.7 profiles and you probably just forgot to start the CI jobs.
I'm positive it builds on all jobs including JDK7 (just without module-info.java).

bmarwell:
Oohhhh I see what you mean now.
I wrote "requires JDK8 for release".

But it's only for building a release with the moditect files included. It still is running on (read carefully): J*RE* 7!

I updated the description to emphasize the difference between a JDK and JRE and build time and runtime. Of course you know, but I put the blame on me not being a native speaker. 🙈

bmarwell:
Huh… I cannot get the CI to work.

Groovy is generating this class `jjwt/impl/target/generated-sources/groovy-stubs/test/io/jsonwebtoken/impl/io/FakeServiceDescriptorClassLoader.java`:

```groovy
public class NoServiceDescriptorClassLoader
        extends java.lang.ClassLoader implements
        groovy.lang.GroovyObject {
    ;

    public NoServiceDescriptorClassLoader
            (java.lang.ClassLoader parent) {
        super((java.lang.String) null, (java.lang.ClassLoader) null);
    }
}
```

... oddly this fails on zulu-11 and temurin-11, but not on IBM Semeru-11.

This is the Semeru Constructor:
```
protected ClassLoader(String classLoaderName, ClassLoader parentLoader) {
	this(checkSecurityPermission(), classLoaderName, parentLoader);
}
```

This for Zulu/Temurin:
```
    protected ClassLoader(String name, ClassLoader parent) {
        this(checkCreateClassLoader(name), name, parent);
    }
```

Will dig into it later.

TriNextDennis:
@bmarwell can you revisit this PR?
I'd highly appreciate module support!

bmarwell:
Only if @lhazlewood approves. Otherwise I'm not investing time here...

lhazlewood:
@bmarwell you said:

> Will dig into it later.

But we didn't hear anything further, so I never spent any further time reviewing since apparently it was still a work in progress.

Additionally, we only use JDK 7 at release time, so this will have to wait until JDK 7 is dropped anyway.

bmarwell:
> Additionally, we only use JDK 7 at release time, so this will have to wait until JDK 7 is dropped anyway.

This is the actual blocker. I forgot the reason you did not wand to have JDK 9, 11 or 17 for build (which can compile to 7 just fine). I think this PR added a profile which enabled compiling on 9+, so the module-info would only have been present when the build would have been done with JDK 9+ (otherwise it was just missing, which is not good).

Anyway, 

> > Will dig into it later.
> 
> But we didn't hear anything further, so I never spent any further time reviewing since apparently it was still a work in progress.

Probably just an update of dependencies as of now. Feel free to ping me if you are open to building with at least JDK 9 (target/release does not matter). Sorry I forgot to look into it.

Will do a quick rebase anyway, but it will include JDK11 for building the JDK 7 release, just as a  PoC. Can change that later.

bmarwell:
New rebased and squashed uses module-info in `src/main/java9`, which is a better approach. This keeps working when just running `test-compile` or `test`, while moditect runs in `package` phase only (modifies the jar) and thus `test` will not working  on the integration test module.

PR #725, opened 2022-04-22
[#687] trim payload to find json object

bmarwell:
Another approach which does not need JSON nor a Pattern.


lhazlewood:
Closing just as a matter of housekeeping since this logic was merged into `master` via #279. Logic here: https://github.com/jwtk/jjwt/blob/fa1e32bf8633f8935814f6d77b1257ba601402ae/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java#L257-L325

Issue #660, opened 2021-04-23
Unclear error message when parsing date claims


remmelt:
```
SecretKey key = Keys.secretKeyFor(SignatureAlgorithm.HS256);

String jws = Jwts.builder()
        .setExpiration(new Date(2000))
        .signWith(key)
        .compact();

Jwts.parserBuilder().setSigningKey(key)
        .setAllowedClockSkewSeconds(1)
        .setClock(() -> new Date(3001))
        .build()
        .parseClaimsJws(jws);
```

The above code will throw an exception, because the token expired. This is great. The exception thrown has a helpful message:
`io.jsonwebtoken.ExpiredJwtException: JWT expired at 1970-01-01T00:00:02Z. Current time: 1970-01-01T00:00:03Z, a difference of 1 milliseconds.  Allowed clock skew: 1000 milliseconds.`

This got me scratching my head... "If the difference is 1ms, and the allowed skew is 1000ms, that should be cool, right?"

Turns out that the difference displayed already subtracted the allowed skew. In my mind it would be more clear to display

`a difference of 1001 milliseconds.  Allowed clock skew: 1000 milliseconds.`

This would make it immediately obvious that the skew is being taken into account.

What do you think?

lhazlewood:
Dunno why I didn't see this until now.  We'll leave this open to see if we can make this clearer.  Thanks for the issue!

PR #746, opened 2022-07-11
JWT expired exception is now shows difference as now - expired

bdemers:
Instead of now + skew - expired

Fixes: #660


lhazlewood (approved):
Didn't see this for some reason.  Sorry for the delayed review!  Please merge :)

PR #750, opened 2022-08-02
Bump jackson dependency to latest

Mats-SX:
This is important to get the latest security fixes from jackson.
Users of jjwt-jackson have to exclude and manually upgrade this dependency otherwise.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,7 +126,7 @@
         <jdk.version>1.7</jdk.version>
         <buildNumber>${user.name}-${maven.build.timestamp}</buildNumber>
 
-        <jackson.version>2.12.7</jackson.version>
+        <jackson.version>2.12.7.1</jackson.version>
         <orgjson.version>20220320</orgjson.version>
         <gson.version>2.9.0</gson.version>
 


Mats-SX:
Seems there's no interest in merging this.

lhazlewood:
@Mats-SX it's not that there's no interest, it's just that it's lower priority than the JWE work that is consuming project time.  I also thought we upped the version since this PR? I'll double check. Thanks for the bump!

Mats-SX:
@lhazlewood Thanks for the reply. Yes, the version has been updated since I issued this pull request, but not to the latest minor version. I will rebase the PR in order to simplify merging it later.

Mats-SX:
I've changed the version update since the upstream origin changed. I now stay on the `2.12` version of jackson. The currently used version is rather new, but I noticed there's a fix version that covers another vulnerability.

lhazlewood:
I didn't realize that the GitHub Actions for this PR never ran.  They're running now.  After they pass without problems, we'll merge.  Thank you!

PR #774, opened 2023-03-23
Updating build to be reproducible

bdemers:
Multiple builds (without changes) will produce the same output (e.g. the hash of the jars will be identical)

> [!NOTE]
> This can be tested by running:
> ```
> mvn clean install && mvn clean verify artifact:compare
>```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,8 @@
 
     <properties>
 
+        <!-- This date is automatically updated by the release and versions plugins -->
+        <project.build.outputTimestamp>2024-02-17T00:00:00Z</project.build.outputTimestamp>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <arguments />
 
@@ -105,6 +107,7 @@
         <gmavenplus.version>1.6.1</gmavenplus.version> <!-- higher version used in jdk8AndLater profile below -->
         <maven.license.version>4.2.rc3</maven.license.version>
         <maven.license.skipExistingHeaders>true</maven.license.skipExistingHeaders>
+        <felix.version>5.1.9</felix.version>
 
         <jdk.version>7</jdk.version>
         <buildNumber>${user.name}-${maven.build.timestamp}</buildNumber>
@@ -466,7 +469,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-shade-plugin</artifactId>
-                    <version>3.2.1</version>
+                    <version>3.2.3</version>
                     <configuration>
                         <shadedClassifierName>deprecated</shadedClassifierName>
                         <shadedArtifactAttached>true</shadedArtifactAttached>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -592,10 +592,39 @@
...
+                            </filesets>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
-                <version>3.3.0</version>
+                <version>5.1.8</version>
```
bdemers, line 627: IIRC, this will fail CI on the older JVMs
PR is marked as draft as we likely want to wait a bit on this one

lhazlewood: I know this is probably counter to reproducibility, but you can set different versions for different JVMs in the `<profiles>` section at the bottom as a version property.  Or have a different plugin configuration entirely.  Just a thought.

bdemers: That could work 🤔 
I've been wanting to try out a different idea too, building with a newer JDK, and using the toolchain option in the Surefire plugin to test with 7.
https://github.com/actions/setup-java/#maven-options

But, I have a feeling that's more work than it's worth, given that support for 7 will be dropped. 🤷

lhazlewood: Yeah, I dunno how much work you'd want to put in for JDK 7.  We won't need it much longer.

lhazlewood:
@bdemers do you want to try this again given all that stuff has been merged to `master`?  If not, please close. 😄

pom.xml
bdemers: I updated using the profile technique used, it's not worth spending more time with 7 😄

bdemers:
We might want to add a CI check in here 🤔 
Running the build twice for the full (non-7) matrix is probably overkill...

Maybe pick a version, like the min supported reproducible version, (zulu-8)? and add another job?
What do you think?

lhazlewood:
> We might want to add a CI check in here 🤔 Running the build twice for the full (non-7) matrix is probably overkill...
> 
> Maybe pick a version, like the min supported reproducible version, (zulu-8)? and add another job? What do you think?

What do you mean run it twice?  I'm not following 😅

bdemers:
> What do you mean run it twice? I'm not following 😅

Sorry about that!
To check of a project/build reproducible, the build is run twice and the output compared,

Basically this:

```sh
mvn clean install && mvn clean verify artifact:compare
```

This would double the run time, which I don't think is worth doing. We _could_ run the above against a single target (potentially even skipping tests, to speed things up), I'm not sure we should worry about this at all in CI right now.  It's something we could check periodically or at release time.


@hboutemy do you have any thoughts/recommendations for projects on how to validate reproducibility via CI?

hboutemy:
> do you have any thoughts/recommendations for projects on how to validate reproducibility via CI?

I recommend NOT trying to validate reproduciblity via CI: there is no reason reproduciblity will change from commit to commit

checking reproducibility by hand is sufficient, and even **necessary** because 2 builds on the same environment may give the same result, but not on 2 different environments (for example if the current directory is put in the output)

then real checking requires ore than just normal CI

pom.xml
```diff
@@ -678,6 +710,7 @@
             <properties>
                 <maven.jar.version>3.2.2</maven.jar.version>
                 <maven.compiler.version>3.8.1</maven.compiler.version>
+                <felix.version>3.5.1</felix.version>
```
hboutemy, line 713: 3.5.1 is known to have some RB issues: why not upgrade to 5.1.9?

lhazlewood: We still support JDK 7, so that's probably it.

bdemers: yeah 😭, when testing with Java 7 we use the older version, when testing with Java 8+ we use the latest.

We should probably use toolchains for testing with Java 7, but... it's getting dropped soon, so this is a temporary hack.

bdemers:
> I recommend NOT trying to validate reproduciblity via CI: there is no reason reproduciblity will change from commit to commit
> 
> checking reproducibility by hand is sufficient, and even **necessary** because 2 builds on the same environment may give the same result, but not on 2 different environments (for example if the current directory is put in the output)
> 
> then real checking requires ore than just normal CI

Thanks @hboutemy, great points!

PR #777, opened 2023-04-20
Use Acsiidoc as README format

bdemers:
To simplify the release process, we can migrate the README to adoc and use variables for things like the project version.

This _should_ simplify release automation in the future.

NOTE: This PR should NOT be merged, but instead should be re-applied after JWE work is complete

Steps:

Install kramdown:

```bash
gem which kramdown-asciidoc
```

Convert README:

```bash
kramdoc README.md
rm README.md
git add README.adoc
```

~~Add project version variable, and update code blocks:
See b0f106a9932fe0662a50376970c084a3e181a6e4~~

NOTE: Code blocks are not filtered/interpolated by default, so the `subs="+attributes"` directive must be added

TODO:
- [x] Confirm all anchors are present
- [x] Update Note/Warning blocks
- [x] Use adoc Table of Contents
- [ ] Visually compare with README.md (side-by-side)


bdemers:
NOTE: There are other features we can get from adoc too, like automatic generation of the table of contents.  That is not part of this PR.
I also didn't verify the results were the same as the markdown version.  This PR is mostly a POC to check what is possible.

lhazlewood:
@bdemers wanna rebase now?

bdemers:
Rebased, (force pushed)

README.adoc
```diff
@@ -1,502 +1,527 @@
...
-If any of those don't appeal to you, no worries! Any help you would like to offer would be 
-appreciated based on the above caveats concerning [contributing pull requests](#contributing-pull-requests). Feel free
-to [discuss or ask questions first](https://github.com/jwtk/jjwt/discussions) if you're not sure. :)
+If any of those don't appeal to you, no worries! Any help you would like to offer would be
+appreciated based on the above caveats concerning <<contributing-pull-requests,contributing pull requests>>. Feel free
+to https://github.com/jwtk/jjwt/discussions[discuss or ask questions first] if you're not sure. :)
 
-<a name="overview"></a>
-## What is a JSON Web Token?
++++<a name="overview">++++++</a>+++
```
lhazlewood, line 337: Does this mean that the anchor is still valid/retained?  I'm not yet familiar with this type of adoc syntax.

lhazlewood (commented):
2 questions:

1. Is this approach 'worth it' to you - i.e. do you feel it provides both a better authoring mechanism as well as reader/viewer experience?
2. Will existing external links to our documentation still work correctly?  I just thought if something references `https://github.com/jwkt/jjwt/README.md#some-anchor`, everything like that would break, no?

README.adoc
bdemers: Yes, both the named and generated anchors should be the same

bdemers:
> 1. Is this approach 'worth it' to you - i.e. do you feel it provides both a better authoring mechanism as well as reader/viewer experience?

IMHO, it provides a better author experience, and it allows the version to be set with a property (and could be automated at release times).  

> 2. Will existing external links to our documentation still work correctly?  I just thought if something references `https://github.com/jwkt/jjwt/README.md#some-anchor`, everything like that would break, no?

Yes, I made a pass through last week, running something like this (browser console), in both GitHub rendered versions (and diffed them):

```js
for (const a of document.anchors) {
    if (a.name.startsWith('user-content-')) {
        console.log(a.name.substring(13, a.name.length))
    }
}
```

It seems GitHub prefixes the named anchors with `user-content-` (I assume to avoid conflicts).
However, using the shortened fragment, e.g. `README.adoc#some-anchor` works too. (likely some JS to handle the URL fragments 🤷),

lhazlewood (approved):
Ok, thanks for the confirmation!  Feel free to merge - I wouldn't want you to have to go through another big merge conflict again in the future.  If we miss any small details now, we can always adjust them later.

PR #794, opened 2023-08-09
Builder setters

lhazlewood:
Changed all old `set`* mutation methods to more modern builder-style methods: e.g  `setSubject(String)` is now `subject(String)`, etc.

This had a better-than-expected impact on improving the readability and cleanliness of the API, making it less verbose, easier to read and more pleasant to work with.


jglick:
FYI this change caused considerable headaches for clients. Needing to replace calls to previously deprecated methods before updating to a new release is routine if irritating, but renaming an interface method makes it impossible to write code which works with both the older and newer versions of the library, leading to reflection as in https://github.com/hub4j/github-api/pull/1727. It would be more polite to do #308 first so that `@Deprecated default` methods could delegate to the new versions, or just keep both versions `abstract` in the `interface` for at least one release if the expected implementations are also in the library.

Issue #799, opened 2023-08-21
Replace duplicate code with a global environment variable


70825:
**Is your feature request related to a problem? Please describe.**

I find duplicate code and it seems like it can be refactored to eliminate the redundancy.

**Describe the solution you'd like**

All jobs are using the same value for the MVN_CMD environment variable in [ci.yml](https://github.com/jwtk/jjwt/blob/master/.github/workflows/ci.yml)

So I think it would be better to use a global environment variable

[AS-IS]

```yaml
jobs:
  oracle:
    env:
      MVN_CMD: ./mvnw --no-transfer-progress -B
  temurin:
    env:
      MVN_CMD: ./mvnw --no-transfer-progress -B
  ...
```

[TO-BE]

```yaml
env:
  MVN_CMD: ./mvnw --no-transfer-progress -B

...
```

**Describe alternatives you've considered**

x

**Additional context**

If you want this changed, can I try this issue?

I've always wanted to contribute directly to the JWT library that I've been using 😀

lhazlewood:
Thanks for the issue!  If you'd like to open a pull request, and the build works, I'd be happy to include your improvement :)

PR #800, opened 2023-08-22
Replace duplicate code with global environment variable

70825:
closed: #799 

I replace duplicate code with global environment variable

reference1 : https://github.com/kubernetes/minikube/blob/master/.github/workflows/pr.yml#L13-L15
reference2 : https://github.com/yorkie-team/yorkie/blob/main/.github/workflows/ci.yml#L9-L10

I'm so glad to be able to contribute to this 😍

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,14 +6,15 @@ on:
     branches:
       - master
 
+env:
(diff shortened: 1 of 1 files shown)


lhazlewood:
Thank you so much for the PR! 🚀

PR #812, opened 2023-09-06
Added regression tests for #365

### LINKED ISSUE #365: Remove the ability to verify asymmetric signatures with PrivateKeys
JJWT's early days allowed this for server-side convenience - in case a server wanted to verify a JWS with the private key it had access to.  This however encourages poor security practices, potentially exposing the private key unnecessarily.

The JWT specification does require public keys to be used for signature validation (only references public key verification).  Also, the JDK's `Signature.initVerify` only accepts `PublicKey` instances, reinforcing this expectation.  Based on all of this, JJWT should follow suit.

While this is not a backwards-incompatible API change (it would be transparent to JJWT users), it could break existing assumed behavior, so such a change should ideally wait until the 1.0 release.

lhazlewood:
Closes #365 

* Added regression tests
* Removed default doDigest implementation from AbstractSecureDigestAlgorithm and put it in DefaultMacAlgorithm (where it should have been) since SignatureAlgorithms override it already

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ enforcement.
  * Convenient and readable [fluent](http://en.wikipedia.org/wiki/Fluent_interface) interfaces, great for IDE 
    auto-completion to write code quickly
  * Fully RFC specification compliant on all implemented functionality, tested against RFC-specified test vectors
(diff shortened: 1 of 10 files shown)


lhazlewood:
Closes #356

PR #836, opened 2023-09-17
Closes #816

### LINKED ISSUE #773: Misleading exception message for missing serializer .jar
As commented by @darioseidl in https://github.com/jwtk/jjwt/issues/573#issuecomment-898606647:

> with 0.11.2 I'm also seeing this error unless I include the Jackson or Gson jar and the error message is misleading as it says that the `jjwt-impl.jar` is missing, but it's actually the `jjwt-jackson.jar` that was missing in my case:
> 
> ```
>     implementation("io.jsonwebtoken:jjwt-api:0.11.2")
>     runtimeOnly("io.jsonwebtoken:jjwt-impl:0.11.2")
> ```
> 
> results in:
> 
> `Unable to find an implementation for interface io.jsonwebtoken.io.Serializer using java.util.ServiceLoader. Ensure you include a backing implementation .jar in the classpath, for example, jjwt-impl.jar, or your own .jar for custom implementations`
> 
> but this works:
> 
> ```
>     implementation("io.jsonwebtoken:jjwt-api:0.11.2")
>     runtimeOnly("io.jsonwebtoken:jjwt-impl:0.11.2")
>     runtimeOnly("io.jsonwebtoken:jjwt-jackson:0.11.2")
> ```

The message "`for example, jjwt-impl.jar,`" should be changed to "`for example, jjwt-jackson.jar or jjwt-gson.jar`" since `jjwt-impl.jar` is always required and does not reflect a serializer implementation .jar as should be indicated.

### LINKED ISSUE #816: Mistake in PrematureJwtException error message
**Describe the bug**
When a `PrematureJwtException` is thrown, the `differenceMillis` is wrong, since it is calculated as the difference between `nbf` and `minTime`, instead of the difference between `nbf` and `nowTime`.
See [DefaultJwtParser.java#L584](https://github.com/jwtk/jjwt/blob/master/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java#L584)
See also the equivalent calculation for expiration time, line 564.

lhazlewood:
Fixed exception message per recommendation.

Also updated expiration message to be clearer/intuitive.

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtParser.java
@@ -671,11 +671,6 @@ public class DefaultJwtParser implements JwtParser {
 
                     long differenceMillis = nowTime - exp.getTime();
 
-                    /*
-                    String msg = "JWT expired at " + expVal + ". Current time: " + nowVal + ", a difference of " +
-                    differenceMillis + " milliseconds.  Allowed clock skew: " +
-                     */
-
                     String msg = "JWT expired " + differenceMillis + " milliseconds ago at " + expVal + ". " +
                             "Current time: " + nowVal + ". Allowed clock skew: " +
                             this.allowedClockSkewMillis + " milliseconds.";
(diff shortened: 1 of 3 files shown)


lhazlewood:
Closed #816 (incorrect message in commit)

PR #839, opened 2023-09-25
Add Java 21 to CI matrix build

bdemers:
* Update groovy patch version to support Java 21.

NOTE: This adds a Maven profile that ups the source/target version to Java 8, but ONLY when building with 21+


lhazlewood (approved):
LGTM! 🚀

PR #874, opened 2023-11-27
Feat refactored code yash

Yola21:


lhazlewood:
@Yola21 what is the purpose of this PR? There is a lot of whitespace-only changes and API code changes that are not backwards compatible.  Was this intended to be opened against JJWT mainline development? Or were you testing?  Thanks!

PR #883, opened 2023-12-25
#60 Add expireAfter functionality to JWT Builder

### LINKED ISSUE #60: Add a convenience expiration setter which takes a duration
I would imagine that when most people are setting an expiration on a JWT, they will be using some configured timeout duration (e.g. 15 minutes).  the ClaimsMutator has a basic setExpiration helper which takes the expiration date.  It would be great if you added a convenience method which took the duration and did the math to compute the final exipration date, something like:

    public T setExpirationAfter(long duration, TimeUnit timeUnit);

Where the impl is something like:

    return setExpiration(new Date(System.currentTimeMillis() + timeUnit.toMillis(duration)));

pveeckhout:
closes #60 

The `expireAfter` method, accepting duration and timeUnit parameters, has been added to the JwtBuilder interface.

This method calculates the JWT expiration date as either the issuedAt time plus the duration or the system current time plus the duration if an issuedAt time hasn't been set.

Additional tests for this feature have been included in `DefaultJwtParserTest.groovy`.


pveeckhout:
Should I have forgotten some additional testing, or any other remarks, please do not hesitate to comment.

lhazlewood:
Awesome, thanks for this!  We can merge it in the context of #308.

Issue #474, opened 2019-07-15
JwtParser custom validators


lhazlewood:
The current `JwtParser` interface at the time of this issue's creation has mutator methods that enable validation of various claims.

The current `require`* methods only enforce that the claim exists and is exactly equal to a specified value.  While sufficient for many use cases, this is not robust enough to handle many other validations that might be suitable for an application.  For example, applications may want to enforce that a claim:

* Simply exists, regardless of its value
* Exists but isn't null
* Exists and equals an exact value
* Exists and matches a particular value
* Exists and is greater than or less than a particular value
* Maybe exists or not, but if it does, adheres to some particular requirement
* etc...

Because of this variability, this issue is being created to track the work required to introduce a new method (or methods) to specify custom validators.  For example:

`enforce(String claimName, Predicate<T> condition)`

(or similar).

Similarly, sometimes validation might require inspection of more than one claim, e.g.

`validate(Predicate<Claims> condition)`

(or similar).

Note that these methods imply Java 8 lambdas, but that's not strictly necessary.  It would save a lot of work/effort however if we could use lambdas for this.  Given that's already core to most JDK applications, it'd be a hard choice to create our own similar interfaces to simulate lambdas, and that just doesn't sound appealing.

Implementation note:  Based on #473 , these methods should be added *only* to the JwtParserBuilder API and not the existing `JwtParser` API to enforce correct API usage semantics compatible with the 1.0 release.

lhazlewood:
We *might* want to allow disabling key length validation per #521.  This violates the specification, so I don't know how I feel about that, especially if #493 gives an alternative to accomplish the same thing if the user _really_ wants to do this.  In other words #493 makes it possible, but it wouldn't be glaringly easy/obvious by just looking at or using a fluent validator API (which I think is a good thing).

PR #884, opened 2023-12-25
#235 Replace usage of java.util.Date

### LINKED ISSUE #235: Java8 jsr310 LocalDateTime

pveeckhout:
Replace usage of java.util.Date with java.time.Instant in JWT claims

This commit replaces the usage of java.util.Date with java.time.Instant in handling JWT claims. This means upgrading from JDK7 to JDK8. This is done for better adherence to ISO 8601 standards and better precision in time-related operations. Classes working with claims have been updated to use the newer Instant class, providing better precision and cross-timezone compatibility. Additionally, overloading convenience methods for setting OffsetDateTime and ZonedDateTime on top of Instant have been added on the JwtBuilder and DefaultJwtBuilder.

Closes #235


pveeckhout:
There are two failing tests:

`io.jsonwebtoken.JwtParserTest#testParseRequireCustomDate_Success`
`io.jsonwebtoken.JwtParserTest#testParseRequireCustomDate_Incorrect_Fail`

the custom Instant claim is not correct after encryption. I do not immediately see a way to fix this. It might be related to a change in `io.jsonwebtoken.impl.DefaultJwtParser#validateExpectedClaims` where now `Instant` claims are treated as temporal claims where it used to be `Date`

lhazlewood:
@pveeckhout Thanks for taking an interest in helping with this!

However, we're not ready to force a JDK 8 upgrade until remaining JWT functionality is finished.  #308 has some additional work in addition to the Java Date/Time changes for #235.  #474 might also come into play instead of the existing builder utility methods as well.  Some design work needs to be done before we can resolve this in earnest.

I'm happy to leave this PR open of course, but you might have to deal with changes/rebases as we make other changes first.  I hope that's ok!

pveeckhout:
@lhazlewood, My pleasure to help! I am using jjwt in my own projects, and figured it might be time to try and contribute back to project.

I will gladly keep this PR up to date with the other incoming changes.

If there any other issue you figure I could help with, feel free to direct me to them.

api/src/main/java/io/jsonwebtoken/JwtBuilder.java
```diff
@@ -523,14 +525,46 @@ public interface JwtBuilder extends ClaimsMutator<JwtBuilder> {
...
      * <blockquote><pre>
-     * {@link #claims()}.{@link ClaimsMutator#expiration(Date) expiration(exp)}.{@link BuilderClaims#and() and()}</pre></blockquote>
+     * {@link #claims()}.{@link ClaimsMutator#expiration(Instant) expiration(exp)}.{@link BuilderClaims#and() and()}</pre></blockquote>
      *
      * @param exp the JWT {@code exp} value or {@code null} to remove the property from the Claims map.
      * @return the builder instance for method chaining.
      */
     @Override
     // for better/targeted JavaDoc
-    JwtBuilder expiration(Date exp);
```
bdemers, line 533: If adding these other wrappers, keeping the ability to wrap a `Date` would help folks, and lessen the burden when migrating

api/src/main/java/io/jsonwebtoken/ClaimsMutator.java
```diff
@@ -124,10 +124,10 @@ public interface ClaimsMutator<T extends ClaimsMutator<T>> {
      * @param exp the JWT {@code exp} value or {@code null} to remove the property from the JSON map.
      * @return the {@code Claims} instance for method chaining.
      * @deprecated since 0.12.0 in favor of the shorter and more modern builder-style named
-     * {@link #expiration(Date)}. This method will be removed before the JJWT 1.0 release.
+     * {@link #expiration(Instant)}. This method will be removed before the JJWT 1.0 release.
      */
     @Deprecated
-    T setExpiration(Date exp);
+    T setExpiration(Instant exp);
```
bdemers, line 130: Keep the old signatures for the deprecated methods

bdemers (changes requested):
Thanks!
I like this, I was playing around with something similar a little while ago.

I worry a little about folks migrating.  
One option would be to add back the old setter methods with `Date` parameters, potentially marking them as deprecated, though. As much as I prefer using `Instant` lots of folks are probably still using `Date`)

This doesn't work for the getters, but we _could_ add deprecated getters, something like `Date getExpirationDate() { return Date.from(expiration) }`
It would still add a bit of burden on developers when updating, not sure if that's worth it. Or if we should do something else.

Thoughts?

api/src/main/java/io/jsonwebtoken/JwtBuilder.java
pveeckhout: personally I would avoid keeping `java.util.Date` around even as a convenience or temporary warpper

This would require a java version increase, and thus mean a major release. In my opinion it would be the ideal time to 'force' the devs to fix the breaking changes done in the major bump.

ofc, I am not the maintainer and will bend to the will of the code owner. just my 2cents

api/src/main/java/io/jsonwebtoken/ClaimsMutator.java
pveeckhout: this data change requires is a major release, I would actually opt to remove these deprecated methods (in another pull request)

I modified these just to ensure it all compiles and not longer requires `java.time.Date`

bdemers: I get that, but the method is marked as deprecated, so that _shouldn't_ be changed.
(this is basically removing a old deprecated method, and adding a new deprecated method)

In general (and when possible) we want to provide an easy upgrade path for users, (e.g. adding an extra setter and deprecating the old one).
That's not possible with getters, (which would be a breaking change).

bdemers:
Another potential addition related to using the newer dates api would be to add a convenance method to set the `nbf`, `exp`, and `iat` together, something like:

```java
JwtBuilder validFor(Duration duration) {
  Instant now = Instant.now(clock);
  return this.expiration(now.plus(duration))
    .notBefore(now)
    .issuedAt(now);
}
```
I'm not sure that name is clear, or even if this idea would provide enough value to folks.

(This doesn't need to be in this PR, or implemented at all, just adding the thought to the discussion)

pveeckhout:
> Another potential addition related to using the newer dates api would be to add a convenance method to set the `nbf`, `exp`, and `iat` together, something like:
> 
> ```java
> JwtBuilder validFor(Duration duration) {
>   Instant now = Instant.now(clock);
>   return this.expiration(now.plus(duration))
>     .notBefore(now)
>     .issuedAt(now);
> }
> ```
> 
> I'm not sure that name is clear, or even if this idea would provide enough value to folks.
> 
> (This doesn't need to be in this PR, or implemented at all, just adding the thought to the discussion)

fair point, these might be added in another PR that also add some convenience methods, like this one: https://github.com/jwtk/jjwt/pull/883

api/src/main/java/io/jsonwebtoken/ClaimsMutator.java
pveeckhout: the deprecated methods haven been reverted to use `java.util.Date`

tayfun94:
Will someone validate and merge this PR in the future? SonarQube cries everytime I open the JwtUtil class regarding java.util.Date :(

PR #891, opened 2024-01-11
Audience single string claim data type retention when copying

### LINKED ISSUE #890: A single audience value gets converted into a set with one entry
**Describe the bug**

In the scenario when we first build claims with `DefaultClaimsBuilder` and then pass them to `JwtBuilder`, a single audience value gets converted somewhere on the way from `String` to `Set<String>`. **This is a breaking change for the APIs, which expect a string audience value.**

**To Reproduce**
1. Build claims:
```java
var claims = new DefaultClaimsBuilder().audience().single("https://login.salesforce.com").build();
```
2. Check the type of the claim value of in Debugger (must be `String`):
image
3. Create a JWT builder with the claims object:
```java
var jwtBuilder = Jwts.builder().claims(claims);
```
4. Check again the type of the claim value in Debugger (see `jwtBuilder`; this time it's `Set<String>`):
image

**Expected behavior**
In case of using `audience().single(...)` the type of the `audience` claim should stay `String` after calling `claims(...)` on the JWT builder.

**Workaround**
Note, if setting the claim directly on the JWT builder, the type is preserved:
```java
var jwtBuilder = Jwts.builder().audience().single("https://login.salesforce.com");
```

lhazlewood:
Ensured a single string `aud` (Audience) claim would be retained (without converting it to a `Set`) when copying/applying a source Claims instance to a destination Claims builder.

Updated CHANGELOG.md accordingly.

Fixes #890.

### DIFF
--- a/impl/src/test/groovy/io/jsonwebtoken/JwtsTest.groovy
+++ b/impl/src/test/groovy/io/jsonwebtoken/JwtsTest.groovy
@@ -23,6 +23,7 @@ import io.jsonwebtoken.impl.lang.Bytes
 import io.jsonwebtoken.impl.lang.Services
 import io.jsonwebtoken.impl.security.*
 import io.jsonwebtoken.io.Decoders
+import io.jsonwebtoken.io.Deserializer
 import io.jsonwebtoken.io.Encoders
 import io.jsonwebtoken.io.Serializer
 import io.jsonwebtoken.lang.Strings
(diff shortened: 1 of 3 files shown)


impl/src/test/groovy/io/jsonwebtoken/JwtsTest.groovy
```diff
@@ -1167,6 +1168,36 @@ class JwtsTest {
...
+        int i = jws.indexOf('.')
+        int j = jws.lastIndexOf('.')
+        def b64 = jws.substring(i, j)
+        def json = Strings.utf8(Decoders.BASE64URL.decode(b64))
+        def deser = Services.loadFirst(Deserializer)
+        def m = deser.deserialize(new StringReader(json)) as Map<String,?>
+
+        assertEquals aud, m.get('aud') // single string value
+    }
+
```
lhazlewood, line 1200: @devdanylo and @arvindkrishnakumar-okta

Please feel free to take a look at this PR and the test case immediately above, especially lines 1184, 1185 and 1198.

Tests pass locally, I'm just waiting for CI to verify.

lhazlewood:
CI passed successfully, merging.

impl/src/test/groovy/io/jsonwebtoken/JwtsTest.groovy
arvindkrishnakumar-okta: LGTM, thanks for the quick response!

devdanylo: Thank you   🤝

PR #892, opened 2024-01-17
Thread-safe ServiceLoader usage

### LINKED ISSUE #873: Thread safety issue with Services class / ServiceLoader
**Describe the bug**
I am integrating jjwt 0.12.3 into some software, and have discovered what I think to be a possible threading issue in how the Services class interacts with ServiceLoader. I'm using parallel JUnit5 tests and I spuriously (haven't isolated a reliable repro) get NoSuchElementException coming out of java.util.ServiceLoader. The documentation for this class says it's not thread safe, and this is usually the answer given when people query about NoSuchElementException coming out of the service loader, like so:

```
java.util.NoSuchElementException
	at java.base/java.util.ServiceLoader$2.next(ServiceLoader.java:1318)
	at java.base/java.util.ServiceLoader$2.next(ServiceLoader.java:1306)
	at java.base/java.util.ServiceLoader$3.next(ServiceLoader.java:1403)
	at io.jsonwebtoken.impl.lang.Services.loadFirst(Services.java:98)
	at io.jsonwebtoken.impl.DefaultJwtBuilder.compact(DefaultJwtBuilder.java:511)
```

Various ways to solve the issue - a ThreadLocal cache of service loaders, or synchronize loadFirst/loadAll, or wrap the usage of each service loader in a scoped synchronized block, maybe something like

```java
    public static <T> T loadFirst(Class<T> spi) {
        Assert.notNull(spi, "Parameter 'spi' must not be null.");

        ServiceLoader<T> serviceLoader = serviceLoader(spi);
        if (serviceLoader != null) {
            synchronized(serviceLoader) {
                return serviceLoader.iterator().next();
            }
        }

        throw new UnavailableImplementationException(spi);
    }
```

It looks like you cache the loaded service for potential re-use in the Builders, but that's an instance variable and since you use a new builder for each token, it's effectively re-searched on the classpath. Another way to solve this issue would be to load service implementations (i.e. the Serializer, Deserializer, CompressionAlgorithm, etc.) once where it makes sense to do so, potentially in a static initializer somewhere, which the JVM would guarantee in a single thread and therefore be safe. It should also be logically OK since this is all classpath stuff that should never change after program start. Something like this would also be good since you won't have a situation where a large number of request threads needing token validation are synchronizing on JJWT internals unwittingly.

So if possible, I'd go for a non-`synchronized` solution.

**To Reproduce**
Generate tokens in parallel. I'm using jjwt-jackson as my serializer.

**Expected behavior**
NoSuchElement is not thrown spuriously during normal operations.

**Screenshots**
Provided text log above

**Workaround**
Specify the Serializer when using the builder to have the `build` method skip service lookup.

Update: I have noticed that DefaultJwtParserBuilder has a subsequent `build` step that probably steps around this issue by passing resolved services from the ParserBuilder to the Parser. You could do the same for the Builder.

lhazlewood:
Blend of pre-0.11.0 behavior that cached implementation instances and post-0.11.0 behavior using the JDK ServiceLoader to find/create instances of an SPI interface.  This change:

- Reinstates the <= 0.10.x behavior of caching application singleton service implementation instances in a thread-safe reference (previously an `AtomicReference`, but in this change, a `ConcurrentMap`).  If an app singleton instance is cached and found, it is returned to be (re)used immediately when requested.  This is ok for JJWT's purposes because all service implementations instances must be thread-safe application singletons by API contract/design, so caching them for repeated use is fine.

- Ensures that only if a service implementation instance is not in the app singleton cache, a new instance is located/created using a new JDK `ServiceLoader` instance, which doesn't require thread-safe considerations since it is used only in a single-threaded model for the short time it is used to discover a service implementation.  This PR/change removes the post-0.11.0 concurrent cache of `ServiceLoader` instances since they themselves are not designed to be thread-safe.

- Ensures that if a `ServiceLoader` discovers an implementation and returns a new instance, that instance is then cached as an application singleton in the aforementioned `ConcurrentMap` for continued reuse.

- Renames `Services#loadFirst` to `Services#get` to more accurately reflect calling expectations:  The fact that any 'loading' via a `ServiceLoader` may occur is not important for `Services` callers, and the previous method name was unnecessarily exposing internal implementation concepts.  This is safe to do in a point release (0.12.3 -> 0.12.4) because the `Services` class and its methods, while public, are in the `impl` module, only to be used internally for JJWT's purpose and never intended to be used by application developers.

- Updates all test methods to use the renamed method accordingly.

Fixes #873

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java
@@ -15,15 +15,14 @@
  */
 package io.jsonwebtoken.impl.lang;
 
+import io.jsonwebtoken.lang.Arrays;
 import io.jsonwebtoken.lang.Assert;
 
-import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.ServiceLoader;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-
-import static io.jsonwebtoken.lang.Collections.arrayToList;
 
 /**
  * Helper class for loading services from the classpath, using a {@link ServiceLoader}. Decouples loading logic for
@@ -31,9 +30,9 @@ import static io.jsonwebtoken.lang.Collections.arrayToList;
  */
 public final class Services {
 
-    private static ConcurrentMap<Class<?>, ServiceLoader<?>> SERVICE_CACHE = new ConcurrentHashMap<>();
+    private static final ConcurrentMap<Class<?>, Object> SERVICES = new ConcurrentHashMap<>();
 
-    private static final List<ClassLoaderAccessor> CLASS_LOADER_ACCESSORS = arrayToList(new ClassLoaderAccessor[] {
+    private static final List<ClassLoaderAccessor> CLASS_LOADER_ACCESSORS = Arrays.asList(new ClassLoaderAccessor[]{
             new ClassLoaderAccessor() {
                 @Override
                 public ClassLoader getClassLoader() {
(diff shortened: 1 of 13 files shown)


impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java
```diff
@@ -54,86 +53,57 @@ public ClassLoader getClassLoader() {
...
+     * @return The first available instance of the service.
+     * @throws UnavailableImplementationException When no implementation of the SPI class can be found.
      */
-    public static <T> T loadFirst(Class<T> spi) {
-        Assert.notNull(spi, "Parameter 'spi' must not be null.");
-
-        ServiceLoader<T> serviceLoader = serviceLoader(spi);
-        if (serviceLoader != null) {
-            return serviceLoader.iterator().next();
+    public static <T> T get(Class<T> spi) {
```
bdemers, line 69: I like how `Services.get(...)` reads!

bdemers (approved):

PR #894, opened 2024-01-17
#893 Disable Jackson ObjectMapper FAIL_ON_UNKNOWN_PROPERTIES Deserialization Feature by default

### LINKED ISSUE #893: Disable Jackson ObjectMapper FAIL_ON_UNKNOWN_PROPERTIES Deserialization Feature
**Problem**
I use `JacksonDeserializer` constructor with `claimTypeMap` argument in order to parse the values of the claims into my custom types. After a while, I remove unnecessary filed from my custom type. After that, I got Error for all old generated valid JWTs (`UnrecognizedPropertyException`).

Throwing `UnrecognizedPropertyException` when json has Additional/Unknown fields is the default behavior of Jackson deserializer. So if the JWT generator service change its Claim Objects and add some fields, I got error until I add those to my custom type (even if I don't need them).

**Solution**
There are two solution for this problem.
1. Using `@JsonIgnoreProperties(ignoreUnknown = true)` in Class level
2. Configure Jackson Object Mapper and disable `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES`(set to `false`).

I know the first solution solve my problem but I think the purpose of using `JacksonDeserializer` constructor with `claimTypeMap` argument is just to parse the values of the claims into given types (and Not Additional/Unknown fields checking).
And also Others like [Spring Boot disable this feature by default](https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html#howto.spring-mvc.customize-jackson-objectmapper)

So we can change the constructor like this:
```
public JacksonDeserializer(Map<String, Class<?>> claimTypeMap) {
    // DO NOT reuse JacksonSerializer.DEFAULT_OBJECT_MAPPER as this could result in sharing the custom deserializer
    // between instances
    this(new ObjectMapper());
    Assert.notNull(claimTypeMap, "Claim type map cannot be null.");
    // register a new Deserializer
    SimpleModule module = new SimpleModule();
    module.addDeserializer(Object.class, new MappedTypeDeserializer(Collections.unmodifiableMap(claimTypeMap)));
    objectMapper.registerModule(module);
    objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
}
```

But maybe there are some people who count on this exception and use it as a validation. So we can overload the constructor like this:
```
public JacksonDeserializer(Map<String, Class<?>> claimTypeMap) {
    this(claimTypeMap, false);
}

public JacksonDeserializer(Map<String, Class<?>> claimTypeMap, boolean failOnUnknownProperties) {
    // DO NOT reuse JacksonSerializer.DEFAULT_OBJECT_MAPPER as this could result in sharing the custom deserializer
    // between instances
    this(new ObjectMapper());
    Assert.notNull(claimTypeMap, "Claim type map cannot be null.");
    // register a new Deserializer
    SimpleModule module = new SimpleModule();
    module.addDeserializer(Object.class, new MappedTypeDeserializer(Collections.unmodifiableMap(claimTypeMap)));
    objectMapper.registerModule(module);
    objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, failOnUnknownProperties);
}
```

amiriahmad72:
Edit `JacksonDeserializer` constructor with `claimTypeMap` argument:
```
public JacksonDeserializer(Map<String, Class<?>> claimTypeMap) {
    this(claimTypeMap, false);
}

public JacksonDeserializer(Map<String, Class<?>> claimTypeMap, boolean failOnUnknownProperties) {
    // DO NOT reuse JacksonSerializer.DEFAULT_OBJECT_MAPPER as this could result in sharing the custom deserializer
    // between instances
    this(new ObjectMapper());
    Assert.notNull(claimTypeMap, "Claim type map cannot be null.");
    // register a new Deserializer
    SimpleModule module = new SimpleModule();
    module.addDeserializer(Object.class, new MappedTypeDeserializer(Collections.unmodifiableMap(claimTypeMap)));
    objectMapper.registerModule(module);
    objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, failOnUnknownProperties);
}
```
in order to disable Jackson `ObjectMapper` `FAIL_ON_UNKNOWN_PROPERTIES` Deserialization Feature by default .

Closes #893

### DIFF
--- a/extensions/jackson/src/main/java/io/jsonwebtoken/jackson/io/JacksonDeserializer.java
+++ b/extensions/jackson/src/main/java/io/jsonwebtoken/jackson/io/JacksonDeserializer.java
@@ -17,6 +17,7 @@ package io.jsonwebtoken.jackson.io;
 
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.deser.std.UntypedObjectDeserializer;
 import com.fasterxml.jackson.databind.module.SimpleModule;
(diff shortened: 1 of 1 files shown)


lhazlewood:
@amiriahmad72 Thanks for this! But oddly enough, it appears we were working on two nearly identical things at the same time. I opened  https://github.com/jwtk/jjwt/pull/895 recently.

After that's merged, would you like to change this PR to work with that code change?  Some cleanup was done that should enable your addition to be done simply without overloading a constructor.

lhazlewood:
@amiriahmad72 also we'll need a test case to assert the behavior works as expected (and remains working for regression purposes on future builds/releases).

amiriahmad72:
Thanks alot @lhazlewood 
 I close this PR and reopen another one that I use your codes and also **add specific test case**

PR #903, opened 2024-01-26
NIST Elliptic Curve JWKs: field element byte array padding

### LINKED ISSUE #901: NIST Elliptic Curve JWK field element octet string padding
**Describe the bug**
According to RFC 7518, Section [6.2.1.2](https://datatracker.ietf.org/doc/html/rfc7518#section-6.2.1.2) and Section [6.2.1.3](https://datatracker.ietf.org/doc/html/rfc7518#section-6.2.1.3), The length of the X and Y string **MUST** be the full size of a coordinate for the curve specified in the `crv` parameter. For example, if the value of `crv` is `P-521`, the octet string must be 66 octets long.

However, the jjwt library 0.12 version violates this requirement. Instead, it removes the leading zero byte and result in 65 octets long.

**To Reproduce**
Here is a (failed) test case:
```java
// NOT A CONTRIBUTION
package io.jsonwebtoken.impl.security

import io.jsonwebtoken.Jwts
import io.jsonwebtoken.impl.DefaultJwsHeader
import io.jsonwebtoken.io.Decoders
import io.jsonwebtoken.security.Jwks
import org.json.JSONObject
import org.junit.Test

import static org.junit.Assert.assertEquals

class RFC7518Section6Dot2Test {

 @Test
 void test() {
 def key = Jwts.SIG.ES512.keyPair().build().getPublic();
 def jwk = Jwks.builder().key(key).build();
 def header = new DefaultJwsHeader(new JSONObject(jwk).toMap())
 def x = Decoders.BASE64URL.decode((String) header.get("x"))
 def y = Decoders.BASE64URL.decode((String) header.get("y"))
 assertEquals(x.size(), 66)      ////////// fail if X has a leading zero byte
 assertEquals(y.size(), 66)      ////////// fail if Y has a leading zero byte
 }

}
```

**Expected behavior**
jjwt implementation shall **NOT** change the bytes for X, Y coordinates.

lhazlewood:
Ensured NIST Elliptic Curve JWK `x`, `y` and `d` byte array values are padded with leading zero (`0x00`) bytes if necessary before Base64URL-encoding per length requirements defined in:

- https://datatracker.ietf.org/doc/html/rfc7518#section-6.2.1.2
- https://datatracker.ietf.org/doc/html/rfc7518#section-6.2.1.3
- https://datatracker.ietf.org/doc/html/rfc7518#section-6.2.2.1

Fixes #901.

### DIFF
--- a/impl/src/test/groovy/io/jsonwebtoken/impl/security/AbstractEcJwkFactoryTest.groovy
+++ b/impl/src/test/groovy/io/jsonwebtoken/impl/security/AbstractEcJwkFactoryTest.groovy
@@ -15,9 +15,15 @@
  */
 package io.jsonwebtoken.impl.security
 
-
+import io.jsonwebtoken.impl.lang.Bytes
+import io.jsonwebtoken.impl.lang.Services
+import io.jsonwebtoken.io.Decoders
+import io.jsonwebtoken.io.Deserializer
+import io.jsonwebtoken.security.Jwks
 import io.jsonwebtoken.security.UnsupportedKeyException
 import org.junit.Test
+
+import java.security.interfaces.ECPrivateKey
 
 import static org.junit.Assert.assertEquals
 import static org.junit.Assert.fail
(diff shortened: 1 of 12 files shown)


impl/src/test/groovy/io/jsonwebtoken/impl/security/AbstractEcJwkFactoryTest.groovy
```diff
@@ -35,4 +41,36 @@ class AbstractEcJwkFactoryTest {
             assertEquals msg, e.getMessage()
         }
     }
+
+    /**
+     * Asserts correct behavior per https://github.com/jwtk/jjwt/issues/901
+     */
+    @Test
+    void fieldElementByteArrayLength() {
```
lhazlewood, line 49: @7ing here's the updated/modified version of your test.  Thanks for your contribution!

7ing: Thank you for working on this issue.

PR #921, opened 2024-02-07
Move common GH Actions setup to seperate file

bdemers:


bdemers:
This doesn't look possible without workarounds:
https://github.com/orgs/community/discussions/11771
That's probably more complex then dumping these commands in a script

PR #922, opened 2024-02-13
DEVE-142: Create patched jjwt impl that allows 1024 keys

johan-roets:


lhazlewood:
Hi there!

Did you intend to open this PR to the JJWT master branch?  It looks to contain a number of company-specific edits that wouldn't be appropriate in our project.  Also, our minimum JDK is still JDK 7 and we can't upgrade that until after some other issues are finished first.

Finally, the RFC mandatory minimum key length for RSA is 2048 bits for security reasons, and we're not willing to support lower because of this (see https://github.com/jwtk/jjwt?tab=readme-ov-file#rsa for more).

But that doesn't mean you can't do so if you wish - you're welcome to copy/paste the relevant parts of the JJWT `RsaSignatureAlgorithm` source code into your own project and then configure it as a custom algorithm, for example:

```java
MyRsaSignatureAlgorithm myAlg = new MyRsaSignatureAlgorithm();

Jwts.builder().signWith(key, myAlg)/* ... */.compact();

// then configure/override that alg for the parser:

Jwts.parser().sig().add(myAlg).and()./* ... */
```

lhazlewood:
After adding my comment above I noticed this PR is a draft.  Hopefully the above information was helpful though!

PR #935, opened 2024-04-16
Allow using GenericSecret for HmacSHA*

mnylen:
Extend the pre-existing check for SUN PKCS11 generic secret to allow all SecretKeys where getAlgorithm() returns "GenericSecret" to bypass the algorithm validation.

This matches at least with AWS CloudHSM JCE provider, but likely others as well.

Relates to discussion https://github.com/jwtk/jjwt/discussions/934

Let me know if more tests are needed. I didn't find a test that would've tested the existing Sun PKCS#11 generic secret case, and not sure how to test that when the class doesn't exist in the classpath.

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/security/KeysBridge.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/security/KeysBridge.java
@@ -36,6 +36,7 @@ public final class KeysBridge {
 
     private static final String SUNPKCS11_GENERIC_SECRET_CLASSNAME = "sun.security.pkcs11.P11Key$P11SecretKey";
     private static final String SUNPKCS11_GENERIC_SECRET_ALGNAME = "Generic Secret"; // https://github.com/openjdk/jdk/blob/4f90abaf17716493bad740dcef76d49f16d69379/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11KeyStore.java#L1292
+    private static final String GENERIC_SECRET_ALGNAME = "GenericSecret"; // AWS CloudHSM JCE provider and possibly other HSMs
 
     // prevent instantiation
     private KeysBridge() {
@@ -95,10 +96,15 @@ public final class KeysBridge {
         return encoded;
     }
 
-    public static boolean isSunPkcs11GenericSecret(Key key) {
-        return key instanceof SecretKey &&
-                key.getClass().getName().equals(SUNPKCS11_GENERIC_SECRET_CLASSNAME) &&
-                SUNPKCS11_GENERIC_SECRET_ALGNAME.equals(key.getAlgorithm());
+    public static boolean isGenericSecret(Key key) {
+        if (!(key instanceof SecretKey)) {
+            return false;
+        } else if (key.getClass().getName().equals(SUNPKCS11_GENERIC_SECRET_CLASSNAME) &&
+                SUNPKCS11_GENERIC_SECRET_ALGNAME.equals(key.getAlgorithm())) {
+            return true;
+        } else {
+            return GENERIC_SECRET_ALGNAME.equals(key.getAlgorithm());
+        }
     }
 
     /**
--- a/impl/src/main/java/io/jsonwebtoken/impl/security/KeysBridge.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/security/KeysBridge.java
@@ -34,9 +34,9 @@ import java.security.interfaces.RSAKey;
 @SuppressWarnings({"unused"}) // reflection bridge class for the io.jsonwebtoken.security.Keys implementation
 public final class KeysBridge {
 
-    private static final String SUNPKCS11_GENERIC_SECRET_CLASSNAME = "sun.security.pkcs11.P11Key$P11SecretKey";
-    private static final String SUNPKCS11_GENERIC_SECRET_ALGNAME = "Generic Secret"; // https://github.com/openjdk/jdk/blob/4f90abaf17716493bad740dcef76d49f16d69379/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11KeyStore.java#L1292
-    private static final String GENERIC_SECRET_ALGNAME = "GenericSecret"; // AWS CloudHSM JCE provider and possibly other HSMs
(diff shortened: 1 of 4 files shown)


impl/src/main/java/io/jsonwebtoken/impl/security/KeysBridge.java
```diff
@@ -36,6 +36,7 @@ public final class KeysBridge {
 
     private static final String SUNPKCS11_GENERIC_SECRET_CLASSNAME = "sun.security.pkcs11.P11Key$P11SecretKey";
     private static final String SUNPKCS11_GENERIC_SECRET_ALGNAME = "Generic Secret"; // https://github.com/openjdk/jdk/blob/4f90abaf17716493bad740dcef76d49f16d69379/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11KeyStore.java#L1292
+    private static final String GENERIC_SECRET_ALGNAME = "GenericSecret";
```
bdemers, line 39: @mnylen Can you add a comment here about where this string comes from.

You have a great description in the conversation, but adding it inline in the code will help future viewers 😄 

@lhazlewood any idea if this should be made more generic somehow?  I'm not sure how often we would come across this type of problem.  (but I'm guessing it was a PITA to troubleshoot?)

mnylensc: I can add the comment. I also agree that it'd be great to be able to have this support for generic secrets a bit more generic, but not sure how to do that. As far as I'm aware, there's no standard for these HSM-backed JCE providers to follow.

One option of course would be to wrap the key with some kind of additional metadata (such as algorithm), and then unwrap it in the algorithms before passing it to the provider. There's existing `ProviderKey` already, but that is unwrapped way before the key hits the algorithm implementations.

mnylen: I can add the comment. I also agree that it'd be great to be able to have this support for generic secrets a bit more generic, but not sure how to do that. As far as I'm aware, there's no standard for these HSM-backed JCE providers to follow.

One option of course would be to wrap the key with some kind of additional metadata (such as algorithm), and then unwrap it in the algorithms before passing it to the provider. There's existing ProviderKey already, but that is unwrapped way before the key hits the algorithm implementations.

Or perhaps add a mechanism for registering the secret class names that should not be treated as strictly in validations.

lhazlewood: @bdemers @mnylen I'm wondering if this should be made to handle both names, something along the lines of:

```java
if (algName.contains("Generic") && algName.contains("Secret")) ....
```
or
```java
if (algName.startsWith("Generic") && algName.endsWith("Secret"))...
```
whichever is faster.  (I'm mostly sure the .startsWith/.endsWith approach is faster)

mnylen: I don't mind doing the change. Wonder if just startsWith("Generic") would be enough though?

impl/src/main/java/io/jsonwebtoken/impl/security/KeysBridge.java
```diff
@@ -95,10 +96,15 @@ public static byte[] findEncoded(Key key) {
         return encoded;
     }
 
-    public static boolean isSunPkcs11GenericSecret(Key key) {
-        return key instanceof SecretKey &&
-                key.getClass().getName().equals(SUNPKCS11_GENERIC_SECRET_CLASSNAME) &&
-                SUNPKCS11_GENERIC_SECRET_ALGNAME.equals(key.getAlgorithm());
+    public static boolean isGenericSecret(Key key) {
```
lhazlewood, line 99: Because this method previously reflected that it was a PKCS11 key (not just secret), please check to ensure that any calling logic does not reflect PKCS11 needs/context.

If not, and all we care about is a generic secret, then we can probably remove the `SUNPKCS11_GENERIC_SECRET_CLASSNAME` check and keep the `isGenericSecret` rename.

impl/src/main/java/io/jsonwebtoken/impl/security/KeysBridge.java
```diff
@@ -95,10 +96,15 @@ public static byte[] findEncoded(Key key) {
...
+        if (!(key instanceof SecretKey)) {
+            return false;
+        } else if (key.getClass().getName().equals(SUNPKCS11_GENERIC_SECRET_CLASSNAME) &&
```
lhazlewood, line 102: Minor nit:

We tend to prefer pre-emptive returns in their own statement to help with readability, so putting the instanceof check in its own if/return is a little nicer than chaining `else if` logic.  For example, I might rewrite this as:

```java
public static isGenericSecret(Key key) {

    if (!(key instanceof SecretKey) {
        return false;
    }

    String algName = Assert.hasText(key.getAlgorithm(), "Key algorithm cannot be null or empty.");
    return algName.startsWith(GENERIC_PREFIX) && algName.endsWith(SECRET_SUFFIX);
}
```

impl/src/test/groovy/io/jsonwebtoken/impl/security/KeysBridgeTest.groovy
```diff
@@ -56,4 +60,46 @@ class KeysBridgeTest {
     void testToStringPassword() {
         testFormattedOutput(new PasswordSpec("foo".toCharArray()))
     }
+
+    @Test
+    void testIsGenericSecret() {
+        def genericSecret = new SecretKey() {
+            @Override
+            String getAlgorithm() {
+                return "GenericSecret" ;
```
lhazlewood, line 69: We'd probably want another test/tests with a space between 'Generic' and 'Secret' if we go with the .startsWith/.endsWith approach.

impl/src/main/java/io/jsonwebtoken/impl/security/KeysBridge.java
mnylen: Yeah this had only one use from the DefaultMacAlgorithm. Will check if the comment there needs updating.

mnylen:
Made the requested changes. Been a bit busy, so it's somewhat slow going. :)

PR #937, opened 2024-04-22
JWE arbitrary content compression

mnylen:
The JWE content was compressed only when claims was used as payload, but when using arbitrary content string/bytes/input stream, the compression was omitted, while keeping the "zip" header field, leading to un-parseable JWE as the decompression failed on non-compressed content.

I couldn't find any evidence that this was intended behavior from the documentation or the JWE spec, so made a fix for this.

See https://github.com/jwtk/jjwt/discussions/936

Note that people might have used a similar workaround as I presented in the discussion to manually compress the payload, so fixing this might result in some code breaking somewhere. Might be good to mention in the changelog.

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
@@ -687,7 +687,7 @@ public class DefaultJwtBuilder implements JwtBuilder {
         Assert.stateNotNull(keyAlgFunction, "KeyAlgorithm function cannot be null.");
         assertPayloadEncoding("JWE");
 
-        InputStream plaintext = toInputStream("JWE Unencoded Payload", content);
+        InputStream plaintext = toInputStream("JWE Payload", content);
 
         //only expose (mutable) JweHeader functionality to KeyAlgorithm instances, not the full headerBuilder
         // (which exposes this JwtBuilder and shouldn't be referenced by KeyAlgorithms):
--- a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
@@ -595,14 +595,8 @@ public class DefaultJwtBuilder implements JwtBuilder {
 
             // Next, b64 extension requires the raw (non-encoded) payload to be included directly in the signing input,
             // so we ensure we have an input stream for that:
-            if (payload.isClaims() || payload.isCompressed()) {
-                ByteArrayOutputStream claimsOut = new ByteArrayOutputStream(8192);
-                writeAndClose("JWS Unencoded Payload", payload, claimsOut);
-                payloadStream = Streams.of(claimsOut.toByteArray());
-            } else {
-                // No claims and not compressed, so just get the direct InputStream:
-                payloadStream = Assert.stateNotNull(payload.toInputStream(), "Payload InputStream cannot be null.");
-            }
+            payloadStream = convertPayloadToInputStream(payload);
+
             if (!payload.isClaims()) {
                 payloadStream = new CountingInputStream(payloadStream); // we'll need to assert if it's empty later
             }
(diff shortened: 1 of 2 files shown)


lhazlewood:
Excellent find!!! I can't believe this was missed, thank you so much 🙏

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
```diff
@@ -693,14 +693,13 @@ private String encrypt(final Payload content, final Key key, final Provider keyP
...
 
-        InputStream plaintext;
+        ByteArrayOutputStream out = new ByteArrayOutputStream(4096);
         if (content.isClaims()) {
-            ByteArrayOutputStream out = new ByteArrayOutputStream(4096);
             writeAndClose("JWE Claims", content, out);
-            plaintext = Streams.of(out.toByteArray());
         } else {
-            plaintext = content.toInputStream();
+            writeAndClose("JWE Content", content, out);
```
lhazlewood, line 700: We have to be careful here: we don't want to write the plaintext Inputstream written to an intermediate output stream (and then converted back into an input stream) requiring a roundtrip that isn't necessary if compression is not enabled.

This logic should look more like that found in the `sign` method:

https://github.com/jwtk/jjwt/blob/c673b76ef5b74ddaf51b4e80fc632ddb62b7e29a/impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java#L598-L605

If possible, it'd be ideal if we can move this logic to a helper method that is shared by both `sign` and `encrypt`

lhazlewood (changes requested):

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
mnylen: Makes sense.

I wonder though if we should also be a bit more clever with the initial byte array size for ByteArrayOutputStream. Seems a bit wasteful to do 8 KB byte array allocation if the content is only few hundred bytes. At least if the payload has byte array / string, we could use the initial payload size as the starting point, as even with compression, we shouldn't (at least usually) go above that size.

mnylen: Done

lhazlewood (changes requested):

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
```diff
@@ -595,14 +595,8 @@ private String sign(final Payload payload, final Key key, final Provider provide
...
             // so we ensure we have an input stream for that:
-            if (payload.isClaims() || payload.isCompressed()) {
-                ByteArrayOutputStream claimsOut = new ByteArrayOutputStream(8192);
-                writeAndClose("JWS Unencoded Payload", payload, claimsOut);
-                payloadStream = Streams.of(claimsOut.toByteArray());
-            } else {
-                // No claims and not compressed, so just get the direct InputStream:
-                payloadStream = Assert.stateNotNull(payload.toInputStream(), "Payload InputStream cannot be null.");
-            }
+            payloadStream = toInputStream("JWS Unencoded Payload", payload);
```
lhazlewood, line 598: I'm sorry I didn't make this clear: the term `unencoded payload` only applies when the `b64` extension is in use, and that's only a JWS option.

So in this particular call (line 598), `JWS Unencoded Payload` is correct because it's only being used within the `} else { // b64` block starting on line 591.

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
```diff
@@ -693,14 +687,7 @@ private String encrypt(final Payload content, final Key key, final Provider keyP
...
 
-        InputStream plaintext;
-        if (content.isClaims()) {
-            ByteArrayOutputStream out = new ByteArrayOutputStream(4096);
-            writeAndClose("JWE Claims", content, out);
-            plaintext = Streams.of(out.toByteArray());
-        } else {
-            plaintext = content.toInputStream();
-        }
+        InputStream plaintext = toInputStream("JWE Unencoded Payload", content);
```
lhazlewood, line 690: And on this line `JWE Payload` would be the appropriate name because there is no `b64`/unencoded option for JWE.

Sorry for the confusion!

lhazlewood (changes requested):
Just a minor change remaining, and then I think we can merge.  Thanks for your continued help!

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
mnylen: Done. And no problem. :)

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
lhazlewood: Looks good! Thank you! 🚀

impl/src/main/java/io/jsonwebtoken/impl/DefaultJwtBuilder.java
lhazlewood: Looks good! Thanks! 🚀

lhazlewood (approved):

Issue #961, opened 2024-07-29
JwtParserBuilder NestedCollection overrides


lhazlewood:
Per https://github.com/jwtk/jjwt/issues/521#issuecomment-2256509105, a nested collection `.add` call will _not_ currently replace an existing algorithm with the same identifer (as the JavaDoc says it should).  This issue represents the work to ensure this behavior works as documented.

For those needing this behavior, until this issue is fixed, the current workaround is to explicitly call `.remove` before calling `.add`: 

```java
var alg = new MyCustomAlgorithm();
Jwts.parser().sig()
        .remove(alg) // <-- 1.
        .add(alg)    // <-- 2.
        .and()// ... etc ...
```

lhazlewood:
@lhy-hoyin Sorry for the delay, I had a minor family emergency and haven't been able to look at this, but I will be able to in about a week.  @bdemers would you be able to look? Thanks for the Hacktobertest work!

bdemers:
@lhy-hoyin Sounds good, please send a pull request!

My suggestion would be to think about the unit tests too.
The proposed functionality is very similar to the current `add` method, though likely may need to handle a `null` item differently.

Let us know when you have a PR ready and we will take a look!

> [!TIP]
> Add `Fixes: #961` as the last line in your commit message, to automatically link it to this issue.

PR #970, opened 2024-10-16
Bugfix: Nested collection replacment

lhy-hoyin:
Fixes: #961

### DIFF
--- a/impl/src/main/java/io/jsonwebtoken/impl/lang/DefaultCollectionMutator.java
+++ b/impl/src/main/java/io/jsonwebtoken/impl/lang/DefaultCollectionMutator.java
@@ -23,6 +23,7 @@ import io.jsonwebtoken.lang.Strings;
 
 import java.util.Collection;
 import java.util.LinkedHashSet;
+import java.util.NoSuchElementException;
 
 public class DefaultCollectionMutator<E, M extends CollectionMutator<E, M>> implements CollectionMutator<E, M> {
 
@@ -44,6 +45,33 @@ public class DefaultCollectionMutator<E, M extends CollectionMutator<E, M>> impl
             throw new IllegalArgumentException(msg);
         }
         return this.collection.add(e);
+    }
+
+    public M replace(E e) {
+        if (Objects.isEmpty(e)) {
+            if (e == null)  throw new NullPointerException("Cannot be null.");
+            else            throw new IllegalArgumentException("Cannot be empty.");
+        }
+
+        // Keep a copy of the original, in case remove/add fails, and need to rollback
+        E old = null;
+        for (E element : this.collection) {
+            if (element.equals(e)) {
+                old = element;
+                break;
+            }
+        }
+
+        if (old == null || !this.collection.contains(e)) {
+            String msg = this.getClass() + " does not contain " + e + ".";
+            throw new NoSuchElementException(msg);
+        }
+
+        if (this.collection.remove(e))
+            if (doAdd(e)) changed(); // removed and add successfully, notify changed()
+            else this.collection.add(old); // remove successfully but add failed, add back old (not considered a change)
+
+        return self();
     }
 
     @Override
--- a/impl/src/test/groovy/io/jsonwebtoken/impl/lang/DefaultCollectionMutatorTest.groovy
+++ b/impl/src/test/groovy/io/jsonwebtoken/impl/lang/DefaultCollectionMutatorTest.groovy
@@ -16,9 +16,13 @@
 package io.jsonwebtoken.impl.lang
 
 import io.jsonwebtoken.Identifiable
+import io.jsonwebtoken.Jwts
 import io.jsonwebtoken.lang.Strings
+import io.jsonwebtoken.security.MacAlgorithm
 import org.junit.Before
 import org.junit.Test
+
+import java.lang.reflect.Constructor
 
 import static org.junit.Assert.*
 
@@ -126,6 +130,35 @@ class DefaultCollectionMutatorTest {
     void removeMissingDoesNotTriggerChange() {
         m.remove('foo') // not in the collection, no change should be registered
         assertEquals 0, changeCount
+    }
+
+    @Test
+    void replace() {
+        Class<?> c = Class.forName("io.jsonwebtoken.impl.security.DefaultMacAlgorithm")
+        Constructor<?> ctor = c.getDeclaredConstructor(String.class, String.class, int.class)
+        ctor.setAccessible(true)
+        MacAlgorithm custom = (MacAlgorithm) ctor.newInstance('HS512', 'HmacSHA512', 80)
+
+        m.add(Jwts.SIG.HS512)
+        m.replace(custom)
+        assertEquals 2, changeCount // replace is count as one change
+        assertEquals 1, m.getCollection().size() // existing is removed as part of replacement
+        assertEquals 80, ((MacAlgorithm) m.getCollection().toArray()[0]).getKeyBitLength()
+    }
+
+    @Test(expected = NoSuchElementException)
+    void replaceMissing() {
+        m.replace('foo')
+    }
+
+    @Test(expected = NullPointerException)
+    void replaceNull() {
+        m.replace(null)
+    }
+
+    @Test(expected = IllegalArgumentException)
+    void replaceEmpty() {
+        m.replace(Strings.EMPTY)
     }
 
     @Test


lhy-hoyin:
I had considered (but eventually decided against) an alternative behavior when `e` cannot be identified to replace something.

Currently, it throws `NoSuchElementException` to indicate that there is nothing "similar" to replace.

As an alternative, when no "similar" element is found, `replace` could behave the same as `doAdd`. Essentially this means "replace if found, otherwise add as new".

I would appreciate thoughts on this alternative.

impl/src/main/java/io/jsonwebtoken/impl/lang/DefaultCollectionMutator.java
```diff
@@ -46,6 +47,33 @@ private boolean doAdd(E e) {
         return this.collection.add(e);
     }
 
+    public M replace(E e) {
+        if (Objects.isEmpty(e)) {
+            if (e == null)  throw new NullPointerException("Cannot be null.");
+            else            throw new IllegalArgumentException("Cannot be empty.");
+        }
```
bdemers, line 54: You can use `Assert.notEmpty()` here, see an example:
https://github.com/jwtk/jjwt/blob/508227cf4d8b59a80df2f7062406b0c47bf624e1/impl/src/main/java/io/jsonwebtoken/impl/lang/CollectionConverter.java#L60

impl/src/main/java/io/jsonwebtoken/impl/lang/DefaultCollectionMutator.java
```diff
@@ -46,6 +47,33 @@ private boolean doAdd(E e) {
...
+
+        // Keep a copy of the original, in case remove/add fails, and need to rollback
+        E old = null;
+        for (E element : this.collection) {
+            if (element.equals(e)) {
+                old = element;
+                break;
+            }
+        }
```
bdemers, line 63: Don't worry about rollback, if the replace fails, this method call should fail.

impl/src/main/java/io/jsonwebtoken/impl/lang/DefaultCollectionMutator.java
```diff
@@ -46,6 +47,33 @@ private boolean doAdd(E e) {
         return this.collection.add(e);
     }
 
+    public M replace(E e) {
```
bdemers, line 50: I would expect this to be something like `public M replace(E old, E new)`, explicitly instruction what should be replaced and with what.

bdemers: However, take note of the `Identifiable` logic above.

impl/src/main/java/io/jsonwebtoken/impl/lang/DefaultCollectionMutator.java
```diff
@@ -46,6 +47,33 @@ private boolean doAdd(E e) {
...
+        }
+
+        if (old == null || !this.collection.contains(e)) {
+            String msg = this.getClass() + " does not contain " + e + ".";
+            throw new NoSuchElementException(msg);
+        }
+
+        if (this.collection.remove(e))
+            if (doAdd(e)) changed(); // removed and add successfully, notify changed()
+            else this.collection.add(old); // remove successfully but add failed, add back old (not considered a change)
```
bdemers, line 72: This is functionally equivalent for non-ordered collections.
In this example, `e.equals(old) == true`, so it _wouldn't_ need to be replaced.  See above note about changing the method signature.

However, for ordered collections (as is defined in the constructor, the `LinkedHashSet`), this would not be a replace. For example, the collection: `[1,2,3]`
replacing `2` with for `A` _should_ result in `[1,A,3]`, but in the above case would result in `[1,3,A]`

In the implementation above calling `replace(2)` would probably end up with the initial collection of `[1,2,3]` being transformed into `[1,3,2]`

impl/src/test/groovy/io/jsonwebtoken/impl/lang/DefaultCollectionMutatorTest.groovy
```diff
@@ -128,6 +132,35 @@ class DefaultCollectionMutatorTest {
         assertEquals 0, changeCount
     }
 
+    @Test
+    void replace() {
```
bdemers, line 136: Assert the contents of the collection as well, not just the number of changes/size.

bdemers (changes requested):
Thanks for opening the PR @lhy-hoyin!

I've left a few overly verbose comments on the your PR, hopefully you find them helpful.

I know you proposed a new `replace` method, and there still might be value in that, I'd suggest starting with a test that described the original problem in #961.

Basically, the docs say that `.add()` should replace the any existing item, most importantly with `Identifiable` objects.  (e.g. where a user wants to replace an algorithm with a different implementation)

Again, thanks for taking the time to look into this issue!

impl/src/main/java/io/jsonwebtoken/impl/lang/DefaultCollectionMutator.java
lhy-hoyin: There is currently no overloaded `Assert.notEmpty()` for `Object`. I shall implement it.

impl/src/test/groovy/io/jsonwebtoken/impl/lang/DefaultCollectionMutatorTest.groovy
lhy-hoyin: I had

> `assertEquals 80, ((MacAlgorithm) m.getCollection().toArray()[0]).getKeyBitLength()`

to check that the original `MacAlgorithm` (with `minKeyBitLength`=512) is indeed replaced with the custom  `MacAlgorithm` (with `minKeyBitLength`=80).

lhy-hoyin:
> `.add()` should replace the any existing item

In the case of `.add(item)` where the item is `Identifiable`, I suppose an existing item would be one with the same `id`.

Otherwise, when item is **not** `Identifiable`, I suppose that an existing item would be determined through the use of `.equals()`.

In other words, the pseudocode would be like
```
add(E e) {
    for (item : collection):
        break when {
            item is Identifiable and item.id == e.id, or
            item.equals(e)
        } -> assign item to old

    if old is empty:
        doAdd(e)
    else:
        replace(old, e)
}
```

bdemers:
> ```
> add(E e) {
>     for (item : collection):
>         break when {
>             item is Identifiable and item.id == e.id, or
>             item.equals(e)
>         } -> assign item to old
> 
>     if old is empty:
>         doAdd(e)
>     else:
>         replace(old, e)
> }
> ```
> 
> Did I interpret correctly?

Yes, possibly with the minor issue of equality.  if `item.equals(e)` there is nothing to do, as `e` is already contained in the collection, and we would not want to trigger the `changed()` method

lhy-hoyin:
I noticed that `DefaultMacAlgorithm` did not override `equals()`, hence its using `CryptoAlgorithm`'s implementation of `equals()`.

In `CryptoAlgorithm`, it has `ID` and `jacName` as attributes.
In `DefaultMacAlgorithm`, apart from inheriting `ID` and `jacName`, it also has it own attribute `minKeyBitLength`.

I would like to implement `equals` in `DefaultMacAlgorithm` (as it simplifies the logic in `replace`), but is unsure if that would be a breaking change (for end users).

```java
    @Override
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }
        if (obj instanceof DefaultMacAlgorithm) {
            DefaultMacAlgorithm other = (DefaultMacAlgorithm) obj;
            return super.equals(obj) && this.minKeyBitLength == other.minKeyBitLength;
        }
        return false;
    }
```
With the above implementation, I am able to pass all tests.

lhy-hoyin:
@bdemers Would you be able to take a look and approve the PR before the end of the month? I am hoping for this PR to contribute to my hacktoberfest progress (which cuts off by the end of the month).

bdemers (changes requested):
Sorry about the delay. We can try to wrap this up this week!

The current `add` flow looks like it needs to iterate over the list twice (once to check if an item matches, and then again to do the replacement.
My suggestion would be to remove the public `replace` method (use a private method if needed), since the main problem is the `add(Identifiable)` use case.

That should reduce the complexity of the method/logic

lhy-hoyin:
@bdemers I have integrated the logic for replacement directly into `add`. This reduced the code complexity. I have also adjusted the code accordingly. 

`replace` is fully removed.

bdemers (approved):
This looks good!
Thank you!

lhazlewood:
@lhy-hoyin sorry for the long delay, but I rebased from `origin/master` and there seems to be a Groovy syntax error in one of your files that is failing the JDK 7 build.  Do you want to take a peek and edit it for a clean build?

impl/src/main/java/io/jsonwebtoken/impl/lang/DefaultCollectionMutator.java
```diff
@@ -48,7 +49,45 @@ private boolean doAdd(E e) {
 
     @Override
     public M add(E e) {
```
lhazlewood, line 51: This is a fairly complex implementation - it almost makes me think we should create a specific abstract subclass for Algorithm/ID-specific logic, e.g. `AlgorithmCollectionMutator` or something like that.

Such a subclass could also be more efficient, potentially using a `Set<String>` or a `Set<String,Identifiable>` to keep track of what IDs exist already instead of iterating over the internal collection.

Finally, there's no method contract that collection element ordering will remain unchanged, so we don't need to perform step 3 - that will implicitly create a performance penalty by saving trailing elements and adding them back in to the end of the collection after replacing the found element.

Thoughts?

lhazlewood (changes requested):
Overall a good PR! Nice work - just have some requested changes and comments.

PR #971, opened 2024-10-17
add properties for graalvm

XenoAmess:
as title.
at least usable in my repo.

### DIFF
--- a/api/src/main/resources/META-INF/native-image/io.jsonwebtoken/jjwt-api/reflect-config.json
+++ b/api/src/main/resources/META-INF/native-image/io.jsonwebtoken/jjwt-api/reflect-config.json
@@ -1,0 +1,209 @@
+[
+  {
+    "name": "io.jsonwebtoken.impl.compression.GzipCompressionCodec",
+    "methods": [
+      {
+        "name": "<init>",
+        "parameterTypes": []
+      }
+    ]
+  },
+  {
+    "name": "io.jsonwebtoken.impl.compression.DeflateCompressionCodec",
+    "methods": [
+      {
+        "name": "<init>",
+        "parameterTypes": []
+      }
+    ]
+  },
+  {
+    "name": "io.jsonwebtoken.impl.security.StandardEncryptionAlgorithms",
+    "methods": [
+      {
+        "name": "<init>",
+        "parameterTypes": []
+      }
+    ]
+  },
+  {
+    "name": "io.jsonwebtoken.impl.security.StandardSecureDigestAlgorithms",
+    "methods": [
+      {
+        "name": "<init>",
+        "parameterTypes": []
+      }
+    ]
+  },
+  {
+    "name": "io.jsonwebtoken.impl.security.StandardKeyAlgorithms",
+    "methods": [
+      {
+        "name": "<init>",
+        "parameterTypes": []
+      }
+    ]
+  },
+  {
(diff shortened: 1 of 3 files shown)


api/src/main/resources/META-INF/native-image/io.jsonwebtoken/jjwt-api/resource-config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "resources": {
+    "includes": [
+      {
+        "pattern": "META-INF/services/io.jsonwebtoken.CompressionCodec"
```
bdemers, line 5: There are a handful of services, though these should _possibly_ live in their corresponding module, and not in `jjwt-api`, thoughts?

```txt
./impl/src/main/resources/META-INF/services/io.jsonwebtoken.CompressionCodec
./extensions/gson/src/main/resources/META-INF/services/io.jsonwebtoken.io.Serializer
./extensions/gson/src/main/resources/META-INF/services/io.jsonwebtoken.io.Deserializer
./extensions/orgjson/src/main/resources/META-INF/services/io.jsonwebtoken.io.Serializer
./extensions/orgjson/src/main/resources/META-INF/services/io.jsonwebtoken.io.Deserializer
./extensions/jackson/src/main/resources/META-INF/services/io.jsonwebtoken.io.Serializer
./extensions/jackson/src/main/resources/META-INF/services/io.jsonwebtoken.io.Deserializer
```

api/src/main/resources/META-INF/native-image/io.jsonwebtoken/jjwt-api/reflect-config.json
```diff
@@ -0,0 +1,209 @@
+[
```
bdemers, line 1: Any thoughts on generating this list at build time?
This is going to be difficult to maintain and track if/when these method signatures change.

Any ideas on creating a test harness? (Extra points if it doesn't require slowly creating a native image in order to test it).

bdemers (commented):
I'm a native-image fan, but this type of thing has always been the hard part.

My main concern is we want to make sure it's _easy_ to maintain as the codebase evolves

api/src/main/resources/META-INF/native-image/io.jsonwebtoken/jjwt-api/resource-config.json
XenoAmess: I put it all in -api for maintaner to maintain them easier, no need to worry about what modification shall look for which json
but yes, logically it should be split into each module

api/src/main/resources/META-INF/native-image/io.jsonwebtoken/jjwt-api/reflect-config.json
XenoAmess: as for generating I have no idea either, maybe I have to make another maven-plugin for it.
for test, unfortunatly I have no idea how to make it test native fastly.

api/src/main/resources/META-INF/native-image/io.jsonwebtoken/jjwt-api/resource-config.json
XenoAmess: I would like to try to write a maven-plugin to auto-generate it, so the two problems shall both be solved.

PR #972, opened 2024-10-18
Made constructor public to allow users their own objectMapper instance

### LINKED ISSUE #914: Make JacksonDeserializer constructor public
The `JacksonDeserializer(ObjectMapper objectMapper, Map<String, Class<?>> claimTypeMap)` constructor was added as a `private` constructor in `0.12.4` due to semver constraints.

This issue tracks the work to make that constructor `public` so devs can still use mapped-type deserialization with their own `ObjectMapper` instance if desired.  This ensures that JJWT does not need to be (regularly) updated any time a particular Jackson deserialization feature/behavior might be preferred - developers can just configure their own `ObjectMapper` with these features/behaviors.

Note that this needs to be done on 0.13.0 or later due to semver.

kesrishubham2510:
Resolves #914


bdemers (approved):
Thanks!
Approved, but we shouldn't merge this until we bump the version for 0.13 (or 1.0), in the near future

PR #989, opened 2025-04-11
Ensured *Builder instances are no longer created with reflection 

### LINKED ISSUE #988: Runtime Builder creation enhancements
Per the discussion in #987, this issue represents work associated with converting `Classes.newInstance` calls to a factory approach that avoids reflection when creating Builder instances.

lhazlewood:
Ensured `*Builder` instances are no longer created with reflection using `Classes.newInstance`.  Instead, `Classes.newInstance` is used to create static/thread-safe Supplier singletons when the api module is loaded, and those singletons will directly instantiate `*Builder` instances at runtime using the `new` keyword`.

Resolves #988.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 ## Release Notes
 
+### 0.12.7
+
+This patch release:
+
+* Improves performance slightly by ensuring all `jjwt-api` utility methods that create `*Builder` instances (`Jwts.builder()`, `Jwts.parserBuilder()`, `Jwks.builder()`, etc) no longer use reflection.
+ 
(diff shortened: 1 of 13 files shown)


lhazlewood:
@slovdahl are you able to build using this PR and test in your application to see if Tomcat plays nice?

lhazlewood:
@slovdahl are you able to build using this PR and test in your application?

slovdahl:
Thanks a lot for working on this! :bow: 

I'll try to get it tested, but a problem right now is that we're still stuck on 0.11.x, and we get some runtime failures when upgrading to 0.12 that we have yet to solve. It might be easier for me to build a 0.11.x version with just this change applied. I'll try to get that done in the coming days.

lhazlewood:
@slovdahl I've decided to merge this because it's a little better for runtime performance in all environments, regardless of Tomcat's particular ClassLoader behavior.  Maybe you can cherry pick this particular PR/commit for testing in your 0.11.x branch - please let us know how it goes!

I'll just add some release note edits to this first and do one more build before merging.

PR #1032, opened 2025-10-28
Added parser for Jackson3

glascaleia:
I created the new parser for jackson3. Regards Giuseppe

PR #1038, opened 2025-12-30
Add Jackson 3 Support

assessmentfoundry:
Adds full Jackson 3.x support while maintaining backward compatibility. See detailed description.


bdemers:
@assessmentfoundry It doesn't look like these tests compile

Can you merge any relevant changes here with #1032?

Issue #582, opened 2020-03-26
isSigned returns true if normal JSON provided or singed with different key


SailReal:
I've two problems with the `boolean isSigned(String jwt)` method: 

The following function call returns *true* if I provide a normal JSON (NOT a singed JWT):

```java
Jwts //
	.parserBuilder() //
	.setSigningKey(getPublicKey()) //
	.build() //
	.isSigned(json);
```

If I change the the method calls to the following:

```java
Jwts //
	.parserBuilder() //
	.setSigningKey(getPublicKey()) //
	.build() //
	.parseClaimsJws(json);
```

a `io.jsonwebtoken.MalformedJwtException: JWT strings must contain exactly 2 period characters. Found: 14` is thrown (which is the expected behavior).  

As `json` a valid JSON is provided, e.g. 
```
{
  "version": "1.5.0-beta1",
  "url": "https://static.cryptomator.org/android/1.5.0-beta1/Cryptomator-1.5.0-beta1.apk",
  "release_notes": "https://static.cryptomator.org/android/1.5.0-beta1/release-notes.md"
}
```

If I provide a valid JWT, signed with a different private key, `isSigned` also returns *true*.

From the doc:
```
* Returns {@code true} if the specified JWT compact string represents a signed JWT (aka a 'JWS'), {@code false}
* otherwise.
* <p>
* <p>Note that if you are reasonably sure that the token is signed, it is more efficient to attempt to
* parse the token (and catching exceptions if necessary) instead of calling this method first before parsing.</p>
```
Do I understand this method in a wrong way? I just want to check if a string is a JWT signed with the corresponding key. In my opinion `isSigned` should return *false* in both cases.

At a different code location I use `parseClaimsJws`, that works great 😍 

As version I use the latest 0.11.1

lhazlewood:
Thanks for the issue!

That method is intended to be used for checking strings reasonably expected to be compact JWT strings - not generic JSON.

Based on the confusion, it is probable that others experience it as well, so it's likely we'll deprecate this method in favor of parsing directly - or perhaps at least change the implementation to delegate to parsing, catch a `MalformedJwtException` and then return `false` as a result, which isn't exactly efficient, so I'm not so sure it's a good idea to propagate its usage.

[...]

PR #1055, opened 2026-06-07
Deprecate JwtParser#isSigned

arpitjain099:
Per the discussion on #582, `JwtParser#isSigned` is a structural-only check slated for removal in 1.0. This marks it `@Deprecated` (interface + impl) with a `@deprecated` Javadoc note steering callers to parse directly and handle the exception, so anyone still using it gets a compiler/IDE nudge before 1.0 removes it.

Notes:
- No behavior change; annotation and Javadoc only.
- Nothing in JJWT's own codebase or tests calls `isSigned`, so there is no internal fallout, and `failOnWarnings` is `false`.
- The Javadoc note is prose (no `@link`) so it is not coupled to a specific replacement method name.

Toward #582 (the actual removal would land in 1.0). Happy to adjust the wording or version note.


lhazlewood:
Thank you!

PR #1066, opened 2026-07-16
Fix empty JSON encrypted content parsing

### LINKED ISSUE #1035: JWE with no "cty" where content accidentally looks like JSON cannot be decrypted with parseEncryptedContent
**Describe the bug**

If I have a content JWE, and it has no "cty", and the payload content happens to be parseable as JSON, then parseEncryptedContent fails with `io.jsonwebtoken.UnsupportedJwtException: Unexpected Claims JWE.`

**To Reproduce**

        Password password = Keys.password("password".toCharArray());
        String jwe = Jwts.builder()
            .content(new byte[] { 0x7B, 0x7D })
            .encryptWith(password, Jwts.KEY.PBES2_HS512_A256KW, Jwts.ENC.A256GCM)
            .compact();
        byte[] payload =  Jwts.parser()
            .decryptWith(password)
            .build()
            .parseEncryptedContent(jwe)
            .getPayload();

This code throws `io.jsonwebtoken.UnsupportedJwtException: Unexpected Claims JWE.`

**Expected behavior**

I expect to get the content, which happens to be the two bytes corresponding to the characters '{' and '}'.

exijn:
Fixes #1035

## Summary
This updates JWT parsing so encrypted content with an empty JSON object byte payload (`{}`) is not automatically treated as Claims unless Claims parsing was explicitly requested.

The previous behavior caused `parseEncryptedContent` to fail with `Unexpected Claims JWE.` when arbitrary encrypted content happened to deserialize as an empty JSON object.

## Changes
- Preserve empty JSON object payloads as content for content parsing paths.
- Keep Claims parsing behavior when Claims are explicitly requested.
- Add a regression test for encrypted content containing `{}` bytes.

## Verification
- `.\mvnw -pl impl -am -Dtest=JwtsTest '-Dsurefire.failIfNoSpecifiedTests=false' test`
- `.\mvnw -pl impl -am -Dtest=JwtParserTest '-Dsurefire.failIfNoSpecifiedTests=false' test`
- `.\mvnw verify`


lhazlewood:
Thank you for the PR!

PR #1075, opened 2026-09-01
Fix JWK key types in README

hanjeun:
Correct the XDH public-key label and EdDSA private-key Java type in the README JWK table.

Verification:
- .\mvnw.cmd clean verify


lhazlewood:
Thank you for the PR!

